PR #11, opened 2019-07-24
Fix secret list offset parameter

vkentta:
Only multiply the offset by limit if both values are positive.

Fixes the usage here:
https://github.com/walmartlabs/concord/blob/434327a1f4b91ff1314441534678fa2d387ad14b/console2/src/components/organisms/SecretDropdown/index.tsx#L97

Currently the above usage offset becomes 1, hiding the first (and possibly only) item in the list.

### DIFF
--- a/console2/src/api/org/secret/index.ts
+++ b/console2/src/api/org/secret/index.ts
@@ -117,7 +117,7 @@ export const list = async (
     limit: number,
     filter?: string
 ): Promise<PaginatedSecretEntries> => {
-    const offsetParam = offset * limit;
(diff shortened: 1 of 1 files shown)


ibodrov:
Thank you for your contribution. It was fixed internally and should be in the public master: https://github.com/walmartlabs/concord/commit/e4c2a85a7a3a84fcfb6299a66fa2352c2d2e37a0

ibodrov:
Merged https://github.com/walmartlabs/concord/commit/495dc3300861db08206eb9a0a5ff1ee6d0655d74

vkentta:
Happy to help :)

Issue #28, opened 2020-06-15
feature: normalize username/userDomain from LDAP Realm before inserting into users table


justenwalker:
https://github.com/walmartlabs/concord/blob/245b41f8e3f68efe648a7d28a707d36a61cd8a88/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserDao.java#L180

https://github.com/walmartlabs/concord/blob/245b41f8e3f68efe648a7d28a707d36a61cd8a88/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserDao.java#L183

Creating api keys fails with `500 Internal Server Error: User not found`  when providing a request

```json
{
  "username": "foo",
  "userDomain": "example.com",
  "userType": "LDAP"
}
```

When looking up the user using `/api/service/console/whoami` they come back as

```json
{ "realm": "ldap", "username": "foo", "userDomain": "Example.com" }
```

The `domain` field in the `users` table has inconsistent case, so normalizing client-side can't be done.

When looking up a user (in LDAP specifically) user@example.com and USER@EXAMPLE.COM are the same user in the same domain, and so they should match. This means either there needs to be some special code for case-insensitive lookup, or data in the users table should be normalized.

As far as I can tell, LDAP DN/Attrs are case-insensitive by default: https://ldapwiki.com/wiki/Distinguished%20Name%20Case%20Sensitivity

I think it would be preferable to normalize the username and userDomain field before inserting from LDAP to mitigate a related issue where two entries for the same account may get inserted.

leading to this error: 

```
java.lang.IllegalArgumentException: Non unique results found for username: 'foo', domain: 'null', type: LDAP
```

ibodrov:
Agreed, DNs are case-insensitive. The LDAP realm should normalize user/domain names before passing them to the rest of the code.

ibodrov:
Taken care in https://github.com/walmartlabs/concord/commit/e436a8b495ecf64a108a553cec9d773b46c25cf5

ibodrov:
Starting from 1.56.0 all user/domain names are forced to lowercase.

PR #29, opened 2020-06-18
Add incasesensitive search & insert

Devjeel:
- [✔] Normalized UserRequest constructor for name & domain

- [✔] Normalized User Data object insertion of name & domain 

- [✔] Normalized UserEntry


Devjeel:
PR for issue - #28

ibodrov:
Thank you, @Devjeel, for your contribution!

The initial assessment that #28 might be a good first ticket,t might not be exactly correct. There is another thing to consider: what to do with the existing data. We probably want to migrate it too. Which requires a [Liquibase changeset](https://github.com/walmartlabs/concord/tree/master/server/db/src/main/resources/com/walmartlabs/concord/server/db).

Another concern is the current usage of `initiator.username` and `currentUser.username` flow variable. Bbefore we can merge it we need to extensively test for situations when someone uses  `${initiator.username == 'someUsernameWithLettersInVariousCase'}` or anything like that. For such users it will be a breaking change.

I'll get back to you once we figure out a migration path.

ibodrov:
https://github.com/walmartlabs/concord/commit/e436a8b495ecf64a108a553cec9d773b46c25cf5 takes care of this.

We decided not to modify data in `CreateUserRequest` and `UserEntry` but handle it on the DAO level.
The commit also takes care of other places where we must normalize user/domain names - e.g. the LDAP data is also queries at the runtime to get the process user information.

Hope that works for you.

PR #31, opened 2020-06-19
Add an option to SecretManager (and it's DAO) to do an upsert

### LINKED ISSUE #30: SecretsManager needs a createOrUpdate method
We're currently using `SecretManager.createBinaryData` - what we really need is `SecretManager.createOrUpdateBinaryData`. We're using SecretManager as a secure storage mechanism only so key recovery isn't an issue. We want to be able to set a secret referenced by name whether it exists or not.

Randgalt:
Add an option to SecretManager (and it's DAO) to do an upsert instead of an insert

Closes #30

How can I test this? Is there a script to set up a local environment?

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/org/secret/SecretDao.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/org/secret/SecretDao.java
@@ -55,6 +55,11 @@ import static org.jooq.impl.DSL.*;
 @Named
 public class SecretDao extends AbstractDao {
 
+    public enum InsertMode {
+        INSERT,
+        UPSERT
+    }
+
     @Inject
     public SecretDao(@MainDB Configuration cfg) {
         super(cfg);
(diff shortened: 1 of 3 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/org/secret/SecretDao.java
```diff
@@ -113,7 +118,9 @@ public UUID insert(DSLContext tx, UUID orgId, UUID projectId, String name, UUID
                         SECRETS.ENCRYPTED_BY,
                         SECRETS.STORE_TYPE,
                         SECRETS.VISIBILITY)
-                .values(name, type.toString(), orgId, projectId, ownerId, encryptedBy.toString(), storeType, visibility.toString())
+                .values(name, type.toString(), orgId, projectId, ownerId, encryptedBy.toString(), storeType, visibility.toString());
+        InsertReturningStep<SecretsRecord> returningStep = (insertMode == InsertMode.UPSERT) ? builder.onDuplicateKeyIgnore() : builder;
```
ibodrov, line 122: Shouldn't it be `onDuplicateKeyUpdate`?

Randgalt: I tried that first, but in fact, this record is created and then the secret data is set in a separate query.  https://github.com/walmartlabs/concord/blob/master/server/impl/src/main/java/com/walmartlabs/concord/server/org/secret/SecretManager.java#L499

PR #41, opened 2020-08-03
concord-server: remove kafka-event-sink from dist

ibodrov:
It was erroneously added in ce3899aec648b89062315e9797ad9ea4c00018d1
cc @brig


brig:
👍

PR #43, opened 2020-08-03
concord-server: store process dependencies in DB

brig:


runtime/loader/src/main/java/com/walmartlabs/concord/process/loader/v2/ConfigurationV2.java
```diff
@@ -27,22 +27,31 @@
...
+
         ObjectMapper om = new ObjectMapper();
         om.registerModule(new Jdk8Module());
         om.registerModule(new JavaTimeModule());
         this.values = om.convertValue(cfg, Map.class);
     }
 
+    @Override
+    public List<String> dependencies() {
+        return cfg.dependencies();
```
ibodrov, line 52: Why not to use the `values` object like in `ConfigurationV1`?
This way we can avoid saving a reference to `ProcessConfiguration` and duplicated data if someone serializes a `ConfigurationV2` instance.

server/db/src/main/resources/com/walmartlabs/concord/server/db/v1.60.0.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet id="1600000" author="ybrigo@gmail.com">
+        <addColumn tableName="PROCESS_QUEUE">
+            <column name="DEPENDENCIES" type="varchar[]">
```
ibodrov, line 9: `text[]` maybe? Like in the previous change sets?

ibodrov:
@brig Do we need a hard limit on the number of dependencies? To prevent someone from putting a million lines into `dependencies` and, subsequently, into the DB.

runtime/loader/src/main/java/com/walmartlabs/concord/process/loader/v2/ConfigurationV2.java
brig: fixed

server/db/src/main/resources/com/walmartlabs/concord/server/db/v1.60.0.xml
brig: fixed

PR #49, opened 2020-08-06
project: more JDK11 compatibility updates

ibodrov:
- add `jdk11` Maven profile;
- fix `Module` imports;
- update JOOQ and Guice versions to get rid of `Illegal reflective access by...` warnings;
- print out path to Java and the current Java version in the Agent's and the Server's start.sh scripts. Useful for debugging.

All tests seems to be happy when I run them locally
```
$ ./mvnw clean install -Pit -Pdocker -Pjdk11
```

cc @brig

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ different systems together using scenarios and plugins created by users.
 
 Dependencies:
 - [Git](https://git-scm.com/) 2.18+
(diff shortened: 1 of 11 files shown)


ibodrov:
Hang on, missed a thing...

ibodrov:
So, for the "Illegal reflective access by..." produced by Guice we will have to wait for https://github.com/google/guice/commit/85e30beafe55551a649025d3c12e831214057412

Other than that, should be g2g.

PR #52, opened 2020-08-10
runtime-v2: remove dependency on v1 SDK

ibodrov:
Make the v2 SDK independent of the v1 SDK:
- create independent interfaces/types for DockerService, SecretService and others.
- update core plugins to support both groups of interfaces/types.

Also, I've removed the `runtime/v2/v1-compat` module. It doesn't work well in practice and I think it will be more productive to focus on keeping clear separation between v1 and v2 types and plugins.


ibodrov:
cc @brig

PR #54, opened 2020-08-10
ignore stack trace log for throw task

brig:


runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/InjectorFactory.java
```diff
@@ -21,7 +21,6 @@
  */
 
 import com.google.inject.*;
-import com.google.inject.Module;
```
ibodrov, line 24: The import is necessary for JDK11 compatibility, i.e. there is [java.lang.Module](https://docs.oracle.com/javase/9/docs/api/java/lang/Module.html) that conflicts with `com.google.inject.Module` if not explicitly specified.

brig: fixed

PR #64, opened 2020-08-13
concord-console: a way to get a link to a specific checkpoint in process logs

brig:
link format:
```
#/process/<instance-id>/log#event-correlation-id
```


pranavparikh:
👍

PR #65, opened 2020-08-15
runtime-v2: common task result structure

brig:


runtime/v2/sdk/src/main/java/com/walmartlabs/concord/runtime/v2/sdk/TaskResult.java
```diff
@@ -0,0 +1,97 @@
...
+        return new TaskResult(true, null, null);
+    }
+
+    public static TaskResult error(String message) {
+        return new TaskResult(false, message, null);
+    }
+
+    private final boolean ok;
+    private final String error;
+    private Map<String, Object> values;
```
ibodrov, line 41: As discussed with @brig, we might want to check if the supplied values are `Serializable`.

PR #71, opened 2020-08-17
runtime-v2: withItems now returns array of results, support for "out" variables in blocks

brig:


ibodrov:
I'm going to postpone the proposed changes - we need to re-iterate on the design some more. The proposed "grouping" of variables in TaskResults works fine for tasks executed in a loop, but not for other constructs that support `withItems` - `call` or `block`.

PR #82, opened 2020-08-24
crypto: create secrets from crypto task runtime-v2

brig:
task params:
```
crypto:
in:
  action: create
  org: <org-name>
  project: <project-name>
  secretName: <secret-name>
  generatePwd: <true|false>
  storePassword: <secret-password>
  visibility: <PUBLIC|PRIVATE>
  keyPair:
  	public: <path-to-public-key>
  	private: <path-to-private-key>
  usernamePassword:
  	username: <username>
  	password: <password>
  data: <path-to-data>
```

### DIFF
--- a/client/src/main/java/com/walmartlabs/concord/client/SecretClient.java
+++ b/client/src/main/java/com/walmartlabs/concord/client/SecretClient.java
@@ -59,6 +59,9 @@ public class SecretClient {
         this.retryInterval = retryInterval;
     }
 
+    /**
+     * Fetches a decrypted Concord secret from the server.
+     */
     public <T extends Secret> T getData(String orgName, String secretName, String password, SecretEntry.TypeEnum expectedType) throws Exception {
         String path = "/api/v1/org/" + orgName + "/secret/" + secretName + "/data";
 
@@ -106,6 +109,9 @@ public class SecretClient {
         }
     }
 
+    /**
+     * Decrypt the provided string using the project's key.
+     */
     public byte[] decryptString(UUID instanceId, byte[] input) throws Exception {
         String path = "/api/v1/process/" + instanceId + "/decrypt";
 
@@ -124,6 +130,21 @@ public class SecretClient {
         }
     }
 
+    /**
+     * Encrypts the provided string using the project's key.
+     */
+    public String encryptString(String orgName, String projectName, String input) throws Exception {
+        return encryptString(null, orgName, projectName, input);
+    }
+
+    /**
+     * Encrypts the provided string using the project's key.
+     *
+     * @deprecated use {@link #encryptString(String, String, String)}.
+     * We don't need {@code instanceId} to encrypt strings anymore, but
+     * we keep this method for backward compatibility.
+     */
+    @Deprecated
     public String encryptString(UUID instanceId, String orgName, String projectName, String input) throws Exception {
         String path = "/api/v1/org/" + orgName + "/project/" + projectName + "/encrypt";
         Map<String, String> headerParams = new HashMap<>();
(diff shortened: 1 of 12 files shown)


plugins/tasks/crypto/pom.xml
```diff
@@ -28,6 +28,15 @@
             <artifactId>concord-runtime-model-v2</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.walmartlabs.concord</groupId>
+            <artifactId>concord-client</artifactId>
```
ibodrov, line 33: provided?

client/src/main/java/com/walmartlabs/concord/client/SecretRequest.java
```diff
@@ -0,0 +1,96 @@
...
+    String storePassword();
+
+    @Nullable
+    SecretEntry.VisibilityEnum visibility();
+
+    @Nullable
+    String project();
+
+    @Nullable
+    Path data();
```
ibodrov, line 52: Not sure if it should be a `Path`, tbh. In some cases it might be desirable to pass the data directly.

client/src/main/java/com/walmartlabs/concord/client/SecretRequest.java
```diff
@@ -0,0 +1,96 @@
...
+
+
+import org.immutables.value.Value;
+
+import javax.annotation.Nullable;
+import java.nio.file.Path;
+
+@Value.Immutable
+@Value.Style(jdkOnly = true)
+public interface SecretRequest {
```
ibodrov, line 31: CreateSecretRequest maybe? It's not like we're using it for requesting secrets from Concord, but rather only to create new secrets.

client/src/main/java/com/walmartlabs/concord/client/SecretClient.java
```diff
@@ -136,6 +138,55 @@ public String encryptString(UUID instanceId, String orgName, String projectName,
         throw new ApiException("Error encrypting string. Status code:" + r.getStatusCode() + " Data: " + r.getData());
     }
 
+    public SecretOperationResponse createSecret(SecretRequest secretRequest) throws IOException {
+        String path = "/api/v1/org/" + secretRequest.org() + "/secret";
+
+        Map<String, Object> params = new HashMap<>();
+        params.put("name", secretRequest.name());
```
ibodrov, line 145: AFAIR, concordclient depends on concord-common and therefore has access to `Constants`. Maybe use those constants there?

plugins/tasks/crypto/src/main/java/com/walmartlabs/concord/plugins/crypto/TaskParams.java
```diff
@@ -0,0 +1,147 @@
...
+import com.walmartlabs.concord.runtime.v2.sdk.Variables;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Map;
+
+public class TaskParams {
+
+    private static final String ACTION_KEY = "action";
+    private static final String NAME_KEY = "name";
```
ibodrov, line 33: `secretName` maybe? Less ambiguous.

runtime/v2/sdk/src/main/java/com/walmartlabs/concord/runtime/v2/sdk/SecretService.java
```diff
@@ -38,6 +47,53 @@
 
     String encryptString(String orgName, String projectName, String value) throws Exception;
 
+    @Value.Immutable
+    @Value.Style(jdkOnly = true)
+    interface SecretOperationResult {
```
ibodrov, line 52: Maybe rename it ot `SecretCreationResult`? Or add a field similar to the Server's `OperationResult` enum? Not sure what's better. In theory, we might want to add methods to update or delete secrets as well and that's when we might want a separate field that indicates the type of operation that was performed.

PR #91, opened 2020-08-26
concord-imports: add support for "dir" imports

ibodrov:
Allows importing of local (external to ${workDir}) directories.
Add Server/Agent configuration parameter to disable the `dir` importer by default.

Example:
```yaml
imports:
  - dir:
      src: "/home/user/concord_stuff"
```

I'll add an IT once testcontainers-concord supports custom Server/Agent configurations in the `DOCKER` mode (https://github.com/concord-workflow/testcontainers-concord/issues/19)


benbroadaway:
Could this be used to snag sensitive info from the container or a directory mounted into the container?

benbroadaway:
Well, apparently that's what `disabledProcessors` default setting is for.

ibodrov:
Yes, it is disabled by default and should only be used in local environment.

PR #118, opened 2020-09-04
concord-console: url anchor and open log segments

d0s056f:
- keep system segment (id: 0) open by default so users can see overall progress; keep others closed
- hyperlinks to individual log segments so users can share/access a segment instead of scrolling


d0s056f:
Screenshot 2020-09-04 at 6 22 29 PM
Screenshot 2020-09-04 at 6 21 53 PM

ibodrov:
LGTM, but does anyone really asking to keep the system segment open by default? I remember there was the opposite sentiment as it rarely contains anything user-specific, other than the process' statuses and unhandled errors.
Maybe we should open it automatically only if there's an error? wdyt @walmartlabs/concord

ibodrov:
I'll merge this PR as it is now, we can change the behaviour later if needed.

ibodrov (approved):

d0s056f:
@ibodrov okay I've made a note of it. @mtkunkel and @jchikatur had a couple of ideas too in yesterday's team sync. let's revisit this later

PR #120, opened 2020-09-06
concord-agent: option to keep workDir after the process ends

ibodrov:
A new configuration option:
```
concord-agent {
  runner {
    persistentWorkDir = "/path/to/dir"
  }
}
```

If specified, Agent will copy all files from the process' working directory
into a subdirectory of `persistentWorkDir` after the process ends (regardless
of the status).

This feature is planned to be used in testcontainers-concord to allow test authors
to inspect the process ${workDir}.

All copied files will be readable by all users (i.e. it sets `0644`
permissions for files and `755` for directories). Should *not* be used in production.

### DIFF
--- a/agent/src/main/java/com/walmartlabs/concord/agent/Agent.java
+++ b/agent/src/main/java/com/walmartlabs/concord/agent/Agent.java
@@ -40,6 +40,7 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 import java.io.IOException;
+import java.nio.file.Files;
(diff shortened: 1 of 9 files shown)


ibodrov:
As an example: https://github.com/concord-workflow/testcontainers-concord/blob/a0737a10515fd606fa65c2ab71c890cc5c8bdf16/testcontainers-concord-core/src/test/java/ca/ibodrov/concord/testcontainers/PersistentWorkdirTest.java

brig (approved):

PR #122, opened 2020-09-07
http-task: allow automatic redirects for POST requests

d0s056f:
post requests in http task should follow page redirects (instead of returning 301) if task parameter `followRedirects` is set to `true` or not provided in the request

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3


d0s056f:
@ibodrov does this require an IT?

benbroadaway:
I think this should an option, but not enabled by default since it's _not_ the default behavior of the HTTP spec.

ibodrov:
> @ibodrov does this require an IT?

Not really, but there are existing unit tests for the task. Might be a good idea to add a unit test for this feature too.

d0s056f:
> I think this should an option, but not enabled by default since it's _not_ the default behavior of the HTTP spec.

@benbroadaway so another flag in the task parameters? something like `allowPostRedirect`?

ibodrov:
There's already `followRedirects` option, which is enabled by default.
While I personally think it should've been `false`, we don't have much choice there.
Let POST use the same option and leave it `true` by default, I guess?

benbroadaway:
Yeah we're kind of in a pickle with the existing `followRedirects` option's behavior versus what should be the default behavior for redirecting POST calls and expanding the total number of parameters. It should be okay to tie it to the `followRedirects` option and make it clear in the documentation that it will follow redirects for each type of call.

d0s056f:
okay. just adding the unit test then

d0s056f:
https://github.com/walmartlabs/concord-website/pull/32

ibodrov:
Getting
```
04:11:50 [INFO ] [ERROR] Tests run: 38, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 8.569 s <<< FAILURE! - in com.walmartlabs.concord.plugins.http.HttpTaskTest
04:11:50 [INFO ] [ERROR] testExecuteGetRequestWithFollowRedirect(com.walmartlabs.concord.plugins.http.HttpTaskTest)  Time elapsed: 0.051 s  <<< ERROR!
04:11:50 [INFO ] org.apache.http.conn.HttpHostConnectException: Connect to localhost:8001 [localhost/127.0.0.1] failed: Connection refused (Connection refused)
04:11:50 [INFO ] Caused by: java.net.ConnectException: Connection refused (Connection refused)
04:11:50 [INFO ] 
04:11:50 [INFO ] [ERROR] testExecutePostRequestWithFollowRedirect(com.walmartlabs.concord.plugins.http.HttpTaskTest)  Time elapsed: 0.037 s  <<< ERROR!
04:11:50 [INFO ] org.apache.http.conn.HttpHostConnectException: Connect to localhost:8001 [localhost/127.0.0.1] failed: Connection refused (Connection refused)
04:11:50 [INFO ] Caused by: java.net.ConnectException: Connection refused (Connection refused)
```

d0s056f:
@ibodrov it's failing because of this stub - https://github.com/walmartlabs/concord/pull/122/files#diff-3900108df33fe2d401f4378aab0de0e4R289
worked on my local as there's a local concord server running on 8001. Should I replace this with a generic endpoint like google .com?

ibodrov:
I think it should be a wiremock endpoint.

d0s056f:
does this work? sorry still learning :)

benbroadaway:
@d0s056f The wiremock-y way to generate the URL is like
```java
        rule.stubFor(post(urlEqualTo("/followRedirectsPost"))
                .willReturn(permanentRedirect(rule.url("/string")))
        );
```

plugins/tasks/http/src/main/java/com/walmartlabs/concord/plugins/http/SimpleHttpClient.java
```diff
@@ -332,6 +333,7 @@ private static CloseableHttpClient createClient(Configuration cfg) throws Except
         return HttpClientBuilder.create()
                 .setConnectionManager(buildConnectionManager())
                 .setDefaultRequestConfig(c.build())
+                .setRedirectStrategy(new LaxRedirectStrategy())
```
benbroadaway, line 336: This should be enabled based on the `followRedirects` option. Something like:
```java
RedirectStrategy redirectStrategy;
if (cfg.isFollowRedirects()) {
    redirectStrategy = new LaxRedirectStrategy();
} else {
    redirectStrategy = DefaultRedirectStrategy.INSTANCE;
}

return HttpClientBuilder.create()
        .setConnectionManager(buildConnectionManager())
        .setDefaultRequestConfig(c.build())
        .setRedirectStrategy(redirectStrategy)
        .build();
```

benbroadaway (changes requested):

plugins/tasks/http/src/main/java/com/walmartlabs/concord/plugins/http/SimpleHttpClient.java
d0s056f: @benbroadaway I have tested this. it works only when followRedirects is true, and doesn't when false. you can check the test cases too

benbroadaway: Aaah so it's already tied to `redirectsEnabled` in the apache `RequestConfig`. Got it.

benbroadaway (approved):

ibodrov:
@concord-ci test it

Issue #160, opened 2020-10-02
Kerberos auth is broken in nested containers.


mtkunkel:
Kinit occurs in the parent container:
```
16:00:20 [INFO ] TGT obtained, expired at 'Thu Oct 01 07:00:20 UTC 2020'
16:00:20 [INFO ] TGT renew at Thu Oct 01 06:59:20 UTC 2020
16:00:21 ANSIBLE: latest: Pulling from walmartlabs/concord-ansible/ansible-docker-pymssql-pyodbc
16:00:21 ANSIBLE: 3c72a8ed6814: Pulling fs layer
16:00:21 ANSIBLE: 46bc21ec4861: Pulling fs layer
16:00:21 ANSIBLE: 3b889782b0ff: Pulling fs layer
```

`KRB5CCNAME` gets dropped in a directory not available in the child container: https://github.com/walmartlabs/concord/blob/master/plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/KerberosAuth.java#L70

`No Kerberos credentials available (default cache: /tmp/concord/prefork6456798604649317788/payload/.concord_tmp/ansible4533715454010284806/tgt-ticket) `

Should the krb5 ticket cache be located in `workdir` like Private Keys (with similar cleanup)?
Should we just mount through the `tempDir` to the child?  Potentially to a different location and re-set `KRB5CCNAME`?

ibodrov:
> Should the krb5 ticket cache be located in workdir like Private Keys (with similar cleanup)?
`/tmp/concord/prefork6456798604649317788/payload` is a path in workDir.

So `/tmp/concord/prefork6456798604649317788/payload/.concord_tmp/ansible4533715454010284806/tgt-ticket` _should_ be available as `/workspace/.concord_tmp/ansible4533715454010284806/tgt-ticket`.

Two potential solutions from the top of my head:
- we can mount the original path in addition to `/workspace`. The directory is available for DIND containers anyway;
- we can make [KerberosAuth](https://github.com/walmartlabs/concord/blob/master/plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/KerberosAuth.java#L70) "DIND-aware" by replacing `workDir` in `tgtCacheFile.toString()` with `/workspace` if it `dockerImage` is specified.

Unfortunately, I don't have an easy way to test it locally. Is there a "Kerberos-in-a-box" kinda thing?

amithkb:
[...]

PR #167, opened 2020-10-07
bug-kerberos-nested-container: change to docker container workspace

amithkb:
https://github.com/walmartlabs/concord/issues/160

Bug when nested containers are used kerberos auth was broken. Fixed by replacing the parent {{workdir}} in tgtCacheFile to container workspace "/workspace"

### DIFF
--- a/plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/KerberosAuth.java
+++ b/plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/KerberosAuth.java
@@ -33,6 +33,7 @@ import sun.security.krb5.internal.ccache.CredentialsCache;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Date;
 import java.util.concurrent.TimeUnit;
 
@@ -70,7 +71,7 @@ public class KerberosAuth implements AnsibleAuth {
     public KerberosAuth enrich(AnsibleEnv env, AnsibleContext context) {
         String dockerImage = getString(context.args(), TaskParams.DOCKER_IMAGE_KEY.getKey());
         if (dockerImage != null) {
-            env.put("KRB5CCNAME", tgtCacheFile.toString().replace(context.workDir().toString(), "/workspace"));
+            env.put("KRB5CCNAME", Paths.get("/workspace").resolve(context.workDir().relativize(tgtCacheFile)).toString());
         }
         else {
             env.put("KRB5CCNAME", tgtCacheFile.toString());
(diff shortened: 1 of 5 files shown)


amithkb:
https://github.com/walmartlabs/concord/issues/160#issuecomment-704980437

plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/KerberosAuth.java
```diff
@@ -66,8 +67,14 @@ public KerberosAuth(String username, String password, Path tmpDir, boolean debug
     }
 
     @Override
-    public KerberosAuth enrich(AnsibleEnv env) {
-        env.put("KRB5CCNAME", tgtCacheFile.toString());
+    public KerberosAuth enrich(AnsibleEnv env, AnsibleContext context) {
+        String dockerImage = getString(context.args(), TaskParams.DOCKER_IMAGE_KEY.getKey());
+        if (dockerImage != null) {
+            env.put("KRB5CCNAME", tgtCacheFile.toString().replace(context.workDir().toString(), "/workspace"));
```
ibodrov, line 73: It might be better to do something like
```java
Paths.get("/workspace").resolve(tgtCacheFile.relativize(workDir))
```
Instead of replacing all `/workspaces` in the path.

amithkb: Can we relativize two uncommon root paths? Because i got `KRB5CCNAME=/workspace/../../..,` when i ran above one

ibodrov: But they should be common. The original `tgtCacheFile` value should start with `workDir`. It's not the case?

amithkb: It should be reverse `Paths.get("/workspace").resolve(context.workDir().relativize(tgtCacheFile))`.

mtkunkel:
Validation of the 1.67.1-SNAPSHOT was successful on the agent as well as inside a nested container. 

I do not have a good way to test this outside of an already Kerberized environment. WIP on automating that test case internally.

RTM once other checks pass and comments are resolved.

ibodrov:
My version of the IT flow doesn't support forked repositories. If `ci/walmart-its` is the same flow I remember - the same applies.

ibodrov (approved):
Local tests are happy.

plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/KerberosAuth.java
ibodrov: Right, I always forget the order in `relativize`.

mtkunkel:
Tests happy here, too.

PR #168, opened 2020-10-07
concord-console: fix column widths on the secret list page

ibodrov:


d0s056f:
@ibodrov now the project column looks a bit too long -

Screenshot 2020-10-07 at 10 03 56 PM

instead, how about removing the collapsible tag from Name column as that could be any size depending on user's entry? looks like below. still not pretty i know -

Screenshot 2020-10-07 at 10 05 51 PM

brig (dismissed):

ibodrov:
@d0s056f yep, that's better.

d0s056f (approved):

PR #181, opened 2020-10-18
runtime-v2: sanitize variables from scripts

brig:


runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/ScriptVariables.java
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =====
+ */
+
+import com.walmartlabs.concord.runtime.v2.sdk.Variables;
+
+import java.util.Map;
+
+public class ScriptVariables implements Variables {
```
ibodrov, line 27: Maybe call it `SanitizedVariables` then? Doesn't seem to have any functionality specific exclusively to scripting.

ibodrov (approved):

PR #185, opened 2020-10-20
metrics: add failed task metrics to collector

amithkb:
server tasks failures are added to metrics (while monitoring tasks table)

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/metrics/FailedTaskError.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/metrics/FailedTaskError.java
@@ -24,11 +24,11 @@ import java.time.OffsetDateTime;
 
 public class FailedTaskError {
 
-    private String taskId;
+    private final String taskId;
 
-    private String taskError;
+    private final String taskError;
 
-    private OffsetDateTime taskErrorAt;
+    private final OffsetDateTime taskErrorAt;
 
     public FailedTaskError(String taskId, String taskError, OffsetDateTime taskErrorAt) {
         this.taskId = taskId;
@@ -40,24 +40,12 @@ public class FailedTaskError {
         return taskId;
     }
 
-    public void setTaskId(String taskId) {
-        this.taskId = taskId;
-    }
-
     public String getTaskError() {
         return taskError;
     }
 
-    public void setTaskError(String taskError) {
-        this.taskError = taskError;
-    }
-
     public OffsetDateTime getTaskErrorAt() {
         return taskErrorAt;
-    }
-
-    public void setTaskErrorAt(OffsetDateTime taskErrorAt) {
-        this.taskErrorAt = taskErrorAt;
     }
 
     @Override
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/metrics/FailedTaskError.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/metrics/FailedTaskError.java
@@ -1,0 +1,71 @@
+package com.walmartlabs.concord.server.metrics;
+
+/*-
+ * *****
+ * Concord
+ * -----
+ * Copyright (C) 2017 - 2018 Walmart Inc.
+ * -----
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


amithkb:
Tests - 

```
# HELP failed_tasks Errors for failed server tasks
# TYPE failed_tasks gauge
failed_tasks{task_id="noderoster/ansible-events-processor",task_error="java.lang.Exception: hh\n	at com.walmartlabs.concord.server.plugins.noderoster.processor.AbstractEventProcessor.performTask(AbstractEventProcessor.java:47)\n	at com.walmartlabs.concord.server.task.TaskScheduler.lambda$startTask$0(TaskScheduler.java:118)\n	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)\n	at java.util.concurrent.FutureTask.run(FutureTask.java:266)\n	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)\n	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)\n	at java.lang.Thread.run(Thread.java:748)\n",task_error_at="2020-10-20T14:36:52.813152+05:30",} 1.0
```

ibodrov:
@concord-ci test it

server/impl/src/main/java/com/walmartlabs/concord/server/metrics/FailedTaskError.java
```diff
@@ -0,0 +1,71 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =====
+ */
+
+import java.time.OffsetDateTime;
+
+public class FailedTaskError {
+
+    private String taskId;
```
ibodrov, line 27: The fields can be `final`.

ibodrov:
In a scenario when a task fails and stays in the failed state for a while - would that mean that the exception and the stacktrace will be saved in each sample or it skips the duplicate values? Can this cause bloat on the Prometheus side?

server/impl/src/main/java/com/walmartlabs/concord/server/metrics/FailedTaskError.java
amithkb: sure

amithkb:
> In a scenario when a task fails and stays in the failed state for a while - would that mean that the exception and the stacktrace will be saved in each sample or it skips the duplicate values? Can this cause bloat on the Prometheus side?

if it is failed for a while it updates each sample in `last_error` and `last_error_at` columns in `tasks` table and also if error recovers later then `last_error_at` will be updated with null with [this](https://github.com/walmartlabs/concord/blob/37de280a6ac25839354b0dd96f73acd55c809da0/server/impl/src/main/java/com/walmartlabs/concord/server/task/SchedulerDao.java#L145). At prometheus side, I think we can check for failure happening continuously for say `x` minutes, we can alert and set frequency to have these checks

ibodrov:
I was mostly worried about blowing up the storage used by Prometheus. If, say, a task stays in failed state for a day or two. Each sample would contain this large string value with the stacktrace and potentially saved lots of times. It's not a problem of this PR, just something to keep in mind.

ibodrov:
@concord-ci test it

amithkb:
> > In a scenario when a task fails and stays in the failed state for a while - would that mean that the exception and the stacktrace will be saved in each sample or it skips the duplicate values? Can this cause bloat on the Prometheus side?
> 
> if it is failed for a while it updates each sample in `last_error` and `last_error_at` columns in `tasks` table and also if error recovers later then `last_error_at` will be updated with null with [this](https://github.com/walmartlabs/concord/blob/37de280a6ac25839354b0dd96f73acd55c809da0/server/impl/src/main/java/com/walmartlabs/concord/server/task/SchedulerDao.java#L145). At prometheus side, I think we can check for failure happening continuously for say `x` minutes, we can alert and set frequency to have these checks

ok. I got it. In that case any other way to implement it, any suggestions?

ibodrov (dismissed):

ibodrov:
> any suggestions?
Not sure, need to look into how metrics are stored exactly and whether there is a deduplication mechanism - i.e. if Prometheus skips a value if it's the same as its previous sample. Keep an eye on the disk space for now :-)

ibodrov (approved):

PR #189, opened 2020-10-20
runtime-v2: eval segment name for all steps

### LINKED ISSUE #187: Variables can not be interpolated into name fields.
For example:
```yaml
- name: Validating client ${item.name}
  call: validate-client
  in:
    staticClient: ${item}
  withItems: ${clients}
```
Outputs `Validating client ${item.name}` instead of something like `Validating client Foo`

Additionally this seems to be inconsistent, as (at lease some) tasks seem to do this properly.

brig:
Fixes #187


ibodrov (approved):

ibodrov:
@concord-ci test it please

ibodrov (approved):

PR #197, opened 2020-10-29
process requirements api: requirement filter + agent operator client …

amithkb:
- adding requirements filter to process requirement api
- update agent operator to add filter during query


k8s/agent-operator/src/main/java/com/walmartlabs/concord/agentoperator/scheduler/Scheduler.java
```diff
@@ -154,7 +154,10 @@ private void updateTargetSize(AgentPoolInstance i) throws IOException {
         }
 
         int queueQueryLimit = i.getResource().getSpec().getQueueQueryLimit();
-        List<ProcessQueueEntry> queueEntries = processQueueClient.query("ENQUEUED", queueQueryLimit);
+        Map<String, Object> queueSelector = i.getResource().getSpec().getQueueSelector();
+        String flavor = mapHasKey(queueSelector,"agent") ? mapHasKey(queueSelector.get("agent"),"flavor") ?
```
brig, line 158: is it possible that `queueSelector.agent` will be not map?
I think it will always a map. 
so you can use `ConfigurationUtils.get(queueSelector, "agent", "flavor")`

server/impl/src/main/java/com/walmartlabs/concord/server/process/queue/FilterUtils.java
```diff
@@ -76,7 +76,7 @@
 
     public static List<ProcessFilter.JsonFilter> parseJson(String paramName, UriInfo uriInfo) {
         return uriInfo.getQueryParameters().entrySet().stream()
-                .filter(e -> isParam(e.getKey(), paramName))
+                .filter(e -> (isParam(e.getKey(), paramName) && isExist(getValue(e))))
```
brig, line 79: may be not send `requeierments` with empty value?
looks like `JsonFilter.value is @Nullable` and value can be null.

amithkb: then, should i have check [here](https://github.com/walmartlabs/concord/blob/45593c7b4f9d8b70b805ba88b0c4d9adf45a890a/k8s/agent-operator/src/main/java/com/walmartlabs/concord/agentoperator/processqueue/ProcessQueueClient.java#L55) for null and construct url?

ibodrov (approved):

PR #198, opened 2020-10-31
runtime-v2: support for parallelWithItems

brig:
```
flows:
  default:
    - task: resultTask
      in:
        result: "${item * 10}"
      out: x
      parallelWithItems:
        - 1
        - 2
        - 3
````
will call `resultTask` in 3 threads.

### CODE REVIEW

File: runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/vm/WithItemsWrapper.java [hunk 1]
```diff
@@ -115,27 +118,97 @@ public void eval(Runtime runtime, State state, ThreadId threadId) {
...
+            frame.push(new JoinCommand(forks.stream().map(Map.Entry::getKey).collect(Collectors.toSet())));
+        }
+    }
+
+    /**
+     * Wraps a command into a loop specified by {@code withItems} option.
+     * Creates a new call frame and keeps the item list, the current item
+     * and the index as frame-local variables.
+     */
+    static class LoopWithItems extends WithItemsWrapper {
```
ibodrov, line 182: Should call those implementations as `SerialWithItems` and `ParallelWithItems` then?

runtime/v2/vm/src/main/java/com/walmartlabs/concord/svm/Frame.java
```diff
@@ -109,6 +113,19 @@ public Serializable getLocal(String k) {
         return Collections.unmodifiableMap(locals);
     }
 
+    public <T> T withLock(Callable<T> c) {
```
ibodrov, line 116: I think you can achieve the same by using `synchronized` on this method instead of a Lock object.

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/vm/WithItemsWrapper.java
brig: yep

runtime/v2/vm/src/main/java/com/walmartlabs/concord/svm/Frame.java
brig: yep

ibodrov (dismissed):

ibodrov (approved):

PR #210, opened 2020-11-04
concord-console: always load all log segments

ibodrov:
Instead of loading a fixed number of segments, load the initial batch
and then keep loading more segments.

It's not proper pagination, but solves the issue of hitting the maximum
(previously hardcoded) number of segments per log.


ibodrov:
I tested it only with a few example flows, but I **think** it should do the trick.

d0s056f:
how are you testing this @ibodrov ?
I have a a v2 flow with 64 segments and it seems to load all as before and with below n/w calls on component mount -

Screenshot 2020-11-05 at 12 09 05 PM

ibodrov:
Alright, I can't completely figure out our hook setup in `ProcessLogActivityV2`, so for now I'm going to set the limit to `-1` and let it load all log segments on every refresh. I'll try to figure out the paginated loading sometime next week.

d0s056f (approved):

brig (approved):

PR #211, opened 2020-11-04
git: support for OAuth tokens

brig:
Now users can create `Single Value` secrets with GH access token and use it to auth to GH repository


brig:
or maybe use `username and password` secrets with empty password for this?

d0s056f (dismissed):

ibodrov (approved):

PR #223, opened 2020-11-09
console: runtime-v2 form links rendering improvements

d0s056f:
form wizard in logs viewer
process status form links (tested with regular + parallel execution)
unique logfile download for each segment


console2/src/components/molecules/LogSegment/index.tsx
```diff
@@ -172,7 +172,7 @@ const LogSegment = ({
                     rel="noopener noreferrer"
                     target="_blank"
                     className="AdditionalAction Last"
-                    data-tooltip="Download Log File"
+                    data-tooltip="Download: InstanceId_SegmentId.log"
```
ibodrov, line 175: Should it be a literal `InstanceId_SegmentId.log` value or it is supposed to be a template?

d0s056f: this is just for the users to understand. the actual template is here https://github.com/walmartlabs/concord/pull/223/files#diff-da1f8800427dd0bdac66a3406695f658b0ae05da7621f3eb2e87c180f2b9a3c6R222

d0s056f:
forgot to update the test cases. labeling wip

d0s056f:
Screenshot 2020-11-10 at 3 06 32 AM

Screenshot 2020-11-10 at 3 06 22 AM

console2/src/components/molecules/LogSegment/index.tsx
d0s056f: I mean we could do it like this, but then users will be confused about the filename - `Download: ${instanceId}_${segmentId}.log`

also, printing the actual values causes the tooltip to become really huge.

console2/src/components/organisms/ProcessLogActivityV2/index.tsx
```diff
@@ -101,14 +107,46 @@ const ProcessLogActivityV2 = ({
         [location]
     );
 
+    const fetchProcessForm = useCallback(async () => {
+        const process = await apiGet(instanceId, ['checkpoints', 'history']);
```
brig, line 111: looks like only `process.status` needed.
maybe just use `processStatus` ?

d0s056f: okay. and I don't need `process` here to check if the process is on? - https://github.com/walmartlabs/concord/pull/223/files#diff-a12794ac01d33384ead17750ed99ae1aee552896dffec038188897bf91e97274R136

brig: yep

d0s056f: done!

ibodrov:
Looks good, however the wizard can fail if you click through the forms too fast.
E.g.

```yaml
configuration:
  runtime: concord-v2

flows:
  default:
    - parallel:
      - form: myFormA
        fields:
          - x: { type: "string" }
      - form: myFormB
        fields:
          - y: { type: "string" }
      - form: myFormC
        fields:
          - z: { type: "string" }

    - log: |
        ${myFormA}
        ${myFormB}
        ${myFormC}
```

- click on the wizard button
- submit the first form
- submit the second form immediately

Fails with
```
14:12:59 [ERROR] Process failed: Process is not suspended, can't resume
java.lang.RuntimeException: Error while updating the process status: Server response: 
	at com.walmartlabs.concord.runtime.v2.runner.remote.DefaultProcessStatusCallback.onRunning(DefaultProcessStatusCallback.java:53)
```

after which all forms are gone (which is weird, looks like it nuked the state, cc @brig). Works perfectly fine if I submit all forms one by one (and waiting for the process to suspend on each step). So, I think, it's just a matter of correctly disabling the UI elements.

So we might need a way to block the wizard until the process is `SUSPENDED`.

We also need to disable form buttons and the wizard button if the process is not `SUSPENDED`.

ibodrov (changes requested):
See above.

d0s056f:
good catch @ibodrov 
wip

ibodrov (approved):

PR #224, opened 2020-11-10
runtime: better instanceId file check

brig:


jchikatur (dismissed):

runtime/v1/impl/src/main/java/com/walmartlabs/concord/runner/Main.java
```diff
@@ -675,6 +683,14 @@ private static void saveLastError(Path baseDir, Throwable t) {
         }
     }
 
+    private static boolean isEmpty(Path path) {
+        if (Files.notExists(path)) {
+            return true;
+        }
+
+        return path.toFile().length() == 0;
```
ibodrov, line 691: https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html ?

runtime/v1/impl/src/main/java/com/walmartlabs/concord/runner/Main.java
```diff
@@ -99,7 +99,7 @@ public void run(RunnerConfiguration runnerCfg, Path baseDir) throws Exception {
         long t1 = System.currentTimeMillis();
 
         Path idPath = baseDir.resolve(Constants.Files.INSTANCE_ID_FILE_NAME);
-        while (!Files.exists(idPath)) {
+        while (isEmpty(idPath)) {
```
ibodrov, line 102: Maybe we should wait until the file is readable *and* it can be parsed as UUID. I.e. loop until the ID parses :-)

ibodrov (approved):

PR #243, opened 2020-11-23
concord-repository: rework the git client

brig:


brig:
@concord-ci test it please

repository/src/main/java/com/walmartlabs/concord/repository/GitCliRepositoryProvider.java
```diff
@@ -51,12 +49,15 @@ public boolean canHandle(String url) {
...
+        return branch != null ? branch : GitConstants.DEFAULT_BRANCH;
     }
 
     @Override
-    public String fetch(String uri, String branchOrNull, String commitId, Secret secret, boolean checkRemoteCommitId, Path dst) {
-        String branch = getBranchOrDefault(branchOrNull);
+    public String fetch(String uri, String branchOrNull, String commitId, Secret secret, Path dst) {
+        String branch = branchOrNull;
+        if (commitId == null && branch == null) {
+            branch = GitConstants.DEFAULT_BRANCH;
```
mtkunkel, line 59: Should this really assume DEFAULT_BRANCH is master?  Soon the default branch may be `main` or whatever is name is picked by git / github. Then we'll have 2 defaults (before the change and after).

Perhaps try and find remote HEAD? `git ls-remote --symref https://github.com/walmartlabs/concord.git HEAD`

brig: I think we can implement this logic in new PR:) we can try get this info from GH, we can make branch name for repo mandatory, we can put default branch name into configuration...

we use this `DEFAULT_BRANCH` constant not only here...

ibodrov (approved):

PR #244, opened 2020-11-24
concord-server,console: remove master branch reference

amithkb:
Github nomenclature change needs to dereference default branch 'master'. 
Update to make repository `branch` as mandatory field and resolve while insertion and update if null


ibodrov:
@concord-ci test it

server/db/src/main/resources/com/walmartlabs/concord/server/db/liquibase.xml
```diff
@@ -91,5 +91,5 @@
     <include file="v1.60.0.xml" relativeToChangelogFile="true"/>
     <include file="v1.66.0.xml" relativeToChangelogFile="true"/>
     <include file="v1.69.0.xml" relativeToChangelogFile="true"/>
-
+    <include file="v1.70.0.xml" relativeToChangelogFile="true"/>
```
ibodrov, line 94: It's going to be 1.75.0+ most likely

amithkb:
@concord-ci test it

amithkb:
@concord-ci test it

ibodrov:
@concord-ci test it

it/server/src/test/java/com/walmartlabs/concord/it/server/ITConstants.java
```diff
@@ -9,9 +9,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
```
ibodrov, line 12: A bit noisy, could you please remove those unnecessary changes?

cli/src/main/java/com/walmartlabs/concord/cli/Run.java
```diff
@@ -100,6 +100,9 @@
     @Option(names = {"-v", "--verbose"}, description = "verbose output")
     boolean verbose = false;
 
+    @Option(names = {"--imports-version/branch"}, description = "default imports version/branch")
```
ibodrov, line 103: I've never seen this format of CLI toggles/parameters with `/` in it. Maybe we can shorten it to `--default-import-version`? Make it more "GNU-ish"? I don't think we'd have other uses for this value, but it sounds more generic and can potentially be useful for other things.

server/dist/src/main/resources/concord-server.conf
```diff
@@ -255,6 +255,9 @@ concord-server {
         disabledProcessors = [
             "dir"
         ]
+
+        # default branch for import
+        defaultBranch = "master"
```
ibodrov, line 260: Perhaps we should put `main` as the default value then?

ibodrov (changes requested):

server/db/src/main/resources/com/walmartlabs/concord/server/db/v1.76.0.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet id="1760000" author="amith.k.b@walmartlabs.com">
+        <sql>
+            update REPOSITORIES set REPO_BRANCH = 'master' where REPO_BRANCH is NULL
```
brig, line 9: `and  commit_id is null`

it/server/src/test/java/com/walmartlabs/concord/it/server/ITConstants.java
amithkb: yes removed :)

cli/src/main/java/com/walmartlabs/concord/cli/Run.java
amithkb: sure

server/db/src/main/resources/com/walmartlabs/concord/server/db/v1.76.0.xml
amithkb: sure

server/dist/src/main/resources/concord-server.conf
amithkb: I think, currently within the organization, the plan to migrate/adoption of `main` is still in progress.  Hence the plan is to keep it configurable and later we will change it to main

amithkb:
@concord-ci test it

console2/public/cfg.js
```diff
@@ -9,5 +9,7 @@ window.concord = {
                 icon: 'github'
             }
         ]
-    }
+    },
+    defaultRepoBranch: 'master'
```
brig, line 13: why we need this default value if branch is mandatory?

amithkb: [here](https://github.com/walmartlabs/concord/blob/baf49a25360152125cc1fc1cc01e0af882982c5b/console2/src/components/molecules/GitHubLink/index.tsx#L79)  - when `commitId` is not present but `path` exists

brig: hmm... commitId or branch is mandatory now. maybe add `branch` as param for GitHubLink?

ibodrov:
@concord-ci test it

console2/public/cfg.js
amithkb: ok removed default value from console config

server/dist/src/main/resources/concord-server.conf
```diff
@@ -418,6 +421,9 @@ concord-server {
 
         sshTimeoutRetryCount = 1
         sshTimeout = "10 minutes"
+
+        # default Git branch
+        defaultBranch = "master"
```
brig, line 426: why we need this default value if branch is mandatory? :)

amithkb: As either commitId or Branch is mandatory. Hence there is still possibility like branch may be null when `repositories` are created by giving only commitId. So i think, we would need this default value @ [1](https://github.com/walmartlabs/concord/blob/6d80f9d09bd10ffd578af5096211800c44111eda/server/impl/src/main/java/com/walmartlabs/concord/server/org/triggers/GithubTriggerEnricher.java#L88), [2](https://github.com/walmartlabs/concord/blob/6d80f9d09bd10ffd578af5096211800c44111eda/server/impl/src/main/java/com/walmartlabs/concord/server/process/pipelines/processors/RepositoryProcessor.java#L111) ... wdyt?

server/dist/src/main/resources/concord-server.conf
benbroadaway: It can be `main` in the repo here, and then customized/overridden in a deployed `server.conf` as necessary.

ibodrov:
@concord-ci test it

ibodrov:
@concord-ci test it

cli/src/main/java/com/walmartlabs/concord/cli/Run.java
```diff
@@ -100,6 +100,9 @@
     @Option(names = {"-v", "--verbose"}, description = "verbose output")
     boolean verbose = false;
 
+    @Option(names = {"--default-import-version"}, description = "default import version or repo branch")
+    String defaultVersion = "master";
```
ibodrov, line 104: If we change the server.conf default value to `main`, we should probably change it here too.
I don't think it's going to cause any real issues for the current users - there aren't that many CLI uses at the moment :-)

brig (approved):

server/dist/src/main/resources/concord-server.conf
amithkb: changed to main

cli/src/main/java/com/walmartlabs/concord/cli/Run.java
amithkb: changed to main

ibodrov (approved):

PR #274, opened 2020-12-16
console: limit form naming pattern

d0s056f:
allow alphanumerics, whitespaces, _ and $ only
works for both runtimes v1 and v2


d0s056f:
Is this okay or I should add it in FormsGrammar.java? Will add/update test cases based on that

ibodrov (changes requested):
Yes, I think we should validate it in Grammar too.

ibodrov:
I feel like this can be closed due to the lack of interest?

PR #278, opened 2020-12-18
concord-tasks: add more methods to the JSON store task

ibodrov:
New methods to check if the record and/or store exists and to insert a
record with automatic store creation.


brig (approved):

brig:
docs: https://github.com/walmartlabs/concord-website/pull/61

PR #279, opened 2020-12-20
runtime-v2: better last error serialization

brig:


runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/vm/SaveLastErrorCommand.java
```diff
@@ -54,14 +56,24 @@ public void eval(Runtime runtime, State state, ThreadId threadId) throws Excepti
...
     }
 
+    private static Map<String, Object> serialize(Exception e) {
+        ObjectMapper mapper = createMapper();
+
+        try {
+            return mapper.convertValue(e, MAP_TYPE);
+        } catch (Exception ex) {
+            // ignore ex
+            return mapper.convertValue(Collections.singletonMap("message", e.getMessage()), MAP_TYPE);
```
ibodrov, line 73: Do we really need `convertValue` here? Can we just do `return Collections.singletonMap("message", e.getMessage())`?

ibodrov: Ah, it's to apply this `IntSequenceGenerator` business, right?

brig: oops. yep, we can just return `singletonMap`

ibodrov (approved):

PR #282, opened 2020-12-22
runtime-v2: use the dependencies classloader for checkpoints

ibodrov:
Explicitly specify the classloader when deserializing checkpoints.

Before the change, checkpoints used the default classloader of the
runner which doesn't contain classes from `dependencies`.

While it is still not recommended to use 3rd party classes as flow
variables, it might be okay for situations when both the server
and the runtime have the same 3rd party JARs on the classpath.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Change log
 
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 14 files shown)


ibodrov:
cc @aczajkowski

brig (approved):

ibodrov:
@concord-ci test it

brig (approved):

PR #291, opened 2021-01-07
server: Allow parameter entry as query parameter

benbroadaway:
Two things make it difficult to compare against a list of values in a json store query
1. JSqlParse chokes on quite a lot of postgres functions and operators
2. Query parameters currently have to be a Map

This is a seemingly simple query that doesn't parse.
```sql
select item_data
from json_store_data
where item_data->'myId' <@ ?::jsonb->'myId'
```

The idea for this change is to continue to use a Map as the given paramters to the REST API call, but add a parameter to specify an entry in the map use actually use as the query parameter when executing the query

```shell
curl -in -X POST -H 'Content-Type: application/json' \
  -d '{"theIds": ["123", "456"]}' \
  'http://localhost:8001/api/v1/org/Default/jsonstore/test-store/query/test-query/exec?paramName=theIds'

[ {
  "status" : "GOOD",
  "myId" : "123"
}, {
  "status" : "GOOD",
  "myId" : "456"
} ]
```

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/org/inventory/InventoryQueryResource.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/org/inventory/InventoryQueryResource.java
@@ -173,7 +173,7 @@ public class InventoryQueryResource implements Resource {
                              @ApiParam @PathParam("queryName") @ConcordKey String queryName,
                              @ApiParam @Valid Map<String, Object> params) {
 
-        return storageQueryResource.exec(orgName, inventoryName, queryName, params);
+        return storageQueryResource.exec(orgName, inventoryName, queryName, null, params);
     }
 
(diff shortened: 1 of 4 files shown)


brig (changes requested):
we need this parser to restrict results with `JSON_STORE_ID`... Maybe we can remove this parser:)

something like this:
- create view for `json_store_data`:
```
create view json_store_data_view_restricted as
select *
from json_store_data
where
    json_store_id = current_setting('jsonStoreQueryExec.json_store_id')::uuid;
```
- replace all `json_store_data` in query with `json_store_data_view_restricted`
- execute in one transaction:
```
set local jsonStoreQueryExec.json_store_id='<real UUID>';

user query;

```

ibodrov:
Btw, that's pretty much how row-level security in PG works. :-)

PR #292, opened 2021-01-07
concord-repository: fix fetching a specific commitId

ibodrov:
When fetching a Git repository, we try to limit the number of remote
refs we fetch.

Before this change fetch might fail if the request specifies
both a branch and commitId - the client was configuring wrong values for
`remote.origin.fetch`.

Steps to reproduce:
- create a Git repository with a default branch + a non default branch with a commit that exist only in the latter;
- configure a repository, use the default branch;
- send a request like so
```
curl -ik -H 'Authorization: ...' -F repoCommitId=$COMMIT_ID_FROM_BRANCH -F org=Default -F project=test -F repo=test http://localhost:8001/api/v1/process
```
- the process should fail with `code: 128, fatal: reference is not a tree: $COMMIT_ID_FROM_BRANCH`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ info popup ([#288](https://github.com/walmartlabs/concord/pull/288)).
 
 ### Changed
 
+- concord-repository: fix Git repository fetching when the request
+specifies the `branchOrTag` and the `commitId` parameters
+simultaneously;
 - runtime-v2: use the standard `AbstractMap.SimpleImmutableEntry`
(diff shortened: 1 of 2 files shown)


mtkunkel (approved):

brig (approved):

ibodrov:
Holding this off as @brig had a better idea how to fix this.

ibodrov:
Closing in favor of https://github.com/walmartlabs/concord/pull/294

PR #303, opened 2021-01-19
ansible: use apache `kerby` instead of `sun.security.*`

brig:


plugins/tasks/ansible/pom.xml
```diff
@@ -12,6 +12,10 @@
     <artifactId>ansible-tasks</artifactId>
     <packaging>takari-jar</packaging>
 
+    <properties>
+        <kerby.version>2.0.1</kerby.version>
```
ibodrov, line 16: Should be in `targetplatform`, I think.

brig: fixed

ibodrov (approved):

PR #318, opened 2021-02-04
concord-server: add pingfed-sso auth plugin

amithkb:
PingIdentity  federation support for concord


brig (dismissed):

amithkb:
Description of the flow:
 - pfed-sso plugin is almost same as iam-sso except changes in REST api contracts used to get `access token`, `id token`  and having an extra api call for logout.
 - separate plugin is created for it and having support for both plugins until the migration because to have feasibility to switch to iam-sso incase of any issues.
 - Once migration is completed we can remove iam-sso module completely

brig (dismissed):

server/plugins/pfed-sso/src/main/java/com/walmartlabs/concord/server/plugins/pfedsso/SsoConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+@Singleton
+public class SsoConfiguration implements Serializable {
+
+    @Inject
+    @Config("sso.pfed.enabled")
+    private boolean enabled;
+
+    @Inject
+    @Config("sso.pfed.priority")
+    private Integer priority;
```
ibodrov, line 42: Can be `int` here, the code treats it as `int` anyway.
+ the same thing in the same class in the iam-sso module.

server/plugins/pfed-sso/src/main/java/com/walmartlabs/concord/server/plugins/pfedsso/SsoClient.java
```diff
@@ -0,0 +1,215 @@
...
+
+                        public void checkServerTrusted(
+                                java.security.cert.X509Certificate[] certs, String authType) {
+                        }
+                    }
+            };
+
+            SSLContext sc = SSLContext.getInstance("SSL");
+            sc.init(null, trustAllCerts, new java.security.SecureRandom());
+            HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());
```
ibodrov, line 85: This looks like a global setting. We probably should not be doing this as it might affect other libraries and/or server plugins.

ibodrov (changes requested):

server/plugins/pfed-sso/src/main/java/com/walmartlabs/concord/server/plugins/pfedsso/SsoConfiguration.java
amithkb: changed to `int`

server/plugins/pfed-sso/src/main/java/com/walmartlabs/concord/server/plugins/pfedsso/SsoClient.java
amithkb: removed as image has trusted cert in it

ibodrov (approved):

PR #320, opened 2021-02-05
replace agent cancel command with just process status update

brig:


server/db/src/main/resources/com/walmartlabs/concord/server/db/liquibase.xml
```diff
@@ -95,6 +95,7 @@
     <include file="v1.76.0.xml" relativeToChangelogFile="true"/>
     <include file="v1.78.0.xml" relativeToChangelogFile="true"/>
     <include file="v1.79.0.xml" relativeToChangelogFile="true"/>
-    <include file="v1.81.0.xml" relativeToChangelogFile="true"/>
+    <include file="v1.80.0.xml" relativeToChangelogFile="true"/>
```
ibodrov, line 98: `v1.80.0.xml` doesn't exist

brig: oops. fixed!

PR #322, opened 2021-02-09
concord-server: use java.time.Duration for all intervals in the config

ibodrov:
Before this change, the Server configuration used a mix of PostgreSQL
interval types and java.time.Duration. Additionally, the existing
Duration properties were converted into milliseconds which causes issues
with large values (e.g. can't set `audit.maxLogAge` interval to more
than `2147483647 ms` or ~24 days.

This PR changes all interval properties to Duration. When the value is
passed to PG it is rounded down to seconds (instead of milliseconds)
which allows values up to ~24000 days.

### DIFF
--- a/server/db/src/main/java/com/walmartlabs/concord/db/PgUtils.java
+++ b/server/db/src/main/java/com/walmartlabs/concord/db/PgUtils.java
@@ -30,11 +30,11 @@ import org.postgresql.util.PSQLException;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Types;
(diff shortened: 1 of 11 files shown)


ibodrov:
@brig I might need help verifying the existing Concord Server configs, in case they use `months` or other units that are not supported by [SimpleConfig#parseDuration](https://github.com/lightbend/config/blob/f92a4ee2f672c8254d87b27831d0903d3d0d6cc2/config/src/main/java/com/typesafe/config/impl/SimpleConfig.java#L735).

brig (approved):

PR #327, opened 2021-02-20
concord-server,pfed-sso: sso login independent of ldap

amithkb:
* SSO authentication is successful after verification of user with AD/LDAP. Hence PR to remove redundant check with LDAP when user logged in through sso.
* Add user info openid endpoint to get user attributes while creating new users
* Add sso redirect on token/session expiration in console

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
@@ -78,12 +78,7 @@ public class LdapUserInfoProvider implements UserInfoProvider {
             // this will result in a 401 response with an empty body anyway
             throw new ConcordApplicationException("Automatic creation of users is disabled.");
         }
-
-//        UserInfo info = getInfo(null, username, userDomain);
-//        if (info == null) {
-//            throw new ConcordApplicationException("User '" + username + "' with domain '" + userDomain + "' not found in LDAP");
-//        }
-
+        
         return userDao.insertOrUpdate(username, userDomain, displayName, email, UserType.LDAP, roles);
     }
 
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
@@ -82,12 +82,6 @@ public class LdapUserInfoProvider implements UserInfoProvider {
         return userDao.insertOrUpdate(username, userDomain, displayName, email, UserType.LDAP, roles);
     }
 
-    @Override
-    public boolean validate(String username, String userDomain) {
-        UserInfo userInfo = getInfo(null, username, userDomain);
-        return userInfo != null;
-    }
-
     private static UserInfo buildInfo(UUID id, LdapPrincipal p) {
         if (p == null) {
             return null;
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
@@ -79,12 +79,18 @@ public class LdapUserInfoProvider implements UserInfoProvider {
             throw new ConcordApplicationException("Automatic creation of users is disabled.");
         }
 
-        UserInfo info = getInfo(null, username, userDomain);
-        if (info == null) {
-            throw new ConcordApplicationException("User '" + username + "' with domain '" + userDomain + "' not found in LDAP");
-        }
+//        UserInfo info = getInfo(null, username, userDomain);
+//        if (info == null) {
+//            throw new ConcordApplicationException("User '" + username + "' with domain '" + userDomain + "' not found in LDAP");
+//        }
 
-        return userDao.insertOrUpdate(info.username(), info.userDomain(), info.displayName(), info.email(), UserType.LDAP, roles);
+        return userDao.insertOrUpdate(username, userDomain, displayName, email, UserType.LDAP, roles);
+    }
+
+    @Override
+    public boolean validate(String username, String userDomain) {
+        UserInfo userInfo = getInfo(null, username, userDomain);
+        return userInfo != null;
     }
 
     private static UserInfo buildInfo(UUID id, LdapPrincipal p) {
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
@@ -20,10 +20,7 @@ package com.walmartlabs.concord.server.security.ldap;
  * =====
  */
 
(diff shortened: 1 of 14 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java
```diff
@@ -88,6 +88,10 @@ public UserManager(UserDao userDao, TeamDao teamDao, AuditLog auditLog, List<Use
             }
         }
 
+        if(!provider.validate(username, userDomain)){
```
ibodrov, line 91: Not sure if I understand the point of this method.
It will result in an additional `UserInfoProvider#getInfo` call [right after we did the first one](https://github.com/walmartlabs/concord/blob/b3e8438fd8f5320d9e16c203b8e64fc18c57428a/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java#L83-L84)?

amithkb: Case for creating new user i,e when user doesn't exist in DB
[LocalUserInfoProvider#getInfo](https://github.com/walmartlabs/concord/blob/4b5a32810a3413e6b4edf382545389c40ec4e34e/server/impl/src/main/java/com/walmartlabs/concord/server/security/internal/LocalUserInfoProvider.java#L51) should give null and [LDAPUserInfoProvider#getInfo](https://github.com/walmartlabs/concord/blob/4b5a32810a3413e6b4edf382545389c40ec4e34e/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java#L64) should give UserInfo for creating a new user.

Since I have removed the secondary validation [#getinfo ](https://github.com/walmartlabs/concord/blob/4b5a32810a3413e6b4edf382545389c40ec4e34e/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java#L82) used to happen in create method, extra `UserInfoProvider#validate` method is enforced. Please let me know if there is any better way to handle this?

server/plugins/pfed-sso/src/main/java/com/walmartlabs/concord/server/plugins/pfedsso/SsoAuthFilter.java
```diff
@@ -68,6 +68,8 @@ public void doFilter(HttpServletRequest request, HttpServletResponse response, F
                 return;
             } else {
                 SsoCookies.removeTokenCookie(response);
+                if (SsoCookies.getRefreshCookie(request) != null)
```
ibodrov, line 71: Missing `{}`

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserResource.java
```diff
@@ -79,6 +79,10 @@ public CreateUserResponse createOrUpdate(@ApiParam @Valid CreateUserRequest req)
 
         UUID id = userManager.getId(username, req.getUserDomain(), type).orElse(null);
         if (id == null) {
+            if (!userManager.validateInfo(username, req.getUserDomain(), type)) {
+                throw new ConcordApplicationException("Validating userinfo failed for user id: "+id+" of type: " + type, Status.BAD_REQUEST);
```
ibodrov, line 83: This is not actionable, it's not going to return anything that would tell the user why the validation failed. Perhaps it should call `UserManager#getOrCreate` instead - assuming the validation there provides the actual error like [the previous implementation did](https://github.com/walmartlabs/concord/blob/master/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java#L84).

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java
ibodrov: I don't understand what we're validating here:
- we call `UserInfoProvider#getInfo` [here](https://github.com/walmartlabs/concord/blob/b3e8438fd8f5320d9e16c203b8e64fc18c57428a/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java#L83);
- then we call `UserInfoProvider#validate` [here](https://github.com/walmartlabs/concord/blob/b3e8438fd8f5320d9e16c203b8e64fc18c57428a/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java#L91);
- `UserInfoProvider#validate` has two implementations - one does nothing and another calls the same `UserInfoProvider#getInfo` method.

Another place where we call `UserInfoProvider#validate` is this [UserManager code](https://github.com/walmartlabs/concord/pull/327/files#diff-0aa5ae1e2a9388a41e0ab9adca08094ab4e5a0ca5ca2c9af845a9a1949cb0164R215).
See my comment there.

Am I missing something or validation does nothing and can be effectively replaced with `UserManager#getOrCreate` in `UserResource`?

amithkb: ok. I refactored a bit. removed `UserInfoProvider#validate` and placed validation in `UserManager#getOrCreate`. Main requirement is to `remove ldap calls when user is created from sso login`

server/plugins/pfed-sso/src/main/java/com/walmartlabs/concord/server/plugins/pfedsso/SsoAuthFilter.java
amithkb: added

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserResource.java
amithkb: implemented as commented

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
```diff
@@ -78,13 +78,8 @@ public UUID create(String username, String userDomain, String displayName, Strin
...
         }
-
-        UserInfo info = getInfo(null, username, userDomain);
-        if (info == null) {
-            throw new ConcordApplicationException("User '" + username + "' with domain '" + userDomain + "' not found in LDAP");
-        }
-
-        return userDao.insertOrUpdate(info.username(), info.userDomain(), info.displayName(), info.email(), UserType.LDAP, roles);
+        
+        return userDao.insertOrUpdate(username, userDomain, displayName, email, UserType.LDAP, roles);
```
ibodrov, line 82: With this change the `UserInfoProvider#create` method becomes redundant - both implementations are the same (just a DAO call). I think we should remove the method and call `userDao.insertOrUpdate(username, userDomain, displayName, email, UserType.LDAP, roles)` in `UserManager`.

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java
ibodrov: :+1:

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
amithkb: done

ibodrov (approved):

PR #332, opened 2021-02-24
runtime-v1/v2: Optionally expose docker daemon

benbroadaway:


agent/src/main/java/com/walmartlabs/concord/agent/cfg/DockerConfiguration.java
```diff
@@ -36,13 +36,15 @@
     private final boolean orphanSweeperEnabled;
     private final long orphanSweeperPeriod;
     private final List<String> extraVolumes;
+    private final Boolean exposeDockerDaemon;
```
ibodrov, line 39: Lower-case `boolean` maybe? I don't think it's a real tri-state value?
The same in other files.

ibodrov (commented):
LGTM % some nitpicking

agent/src/main/java/com/walmartlabs/concord/agent/cfg/DockerConfiguration.java
benbroadaway: Yep, I agree

ibodrov (approved):

PR #333, opened 2021-02-25
concord-server: Get UserInfo from DB on SSO

amithkb:


server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/LdapUserInfoProvider.java
```diff
@@ -51,6 +57,12 @@ public UserType getUserType() {
 
     @Override
     public UserInfo getInfo(UUID id, String username, String userDomain) {
+        /* Get User Data from database when user logged in via SSO */
+        UserPrincipal u = UserPrincipal.getCurrent();
+        if (u != null && u.getRealm().equals(SSO_REALM_NAME)){
```
ibodrov, line 62: This is an abstraction leak. LDAP auth provider should not know anything about other authentication providers.
I don't have a better idea right now, but this is definitely not how it should be.

Same goes for [this](https://github.com/walmartlabs/concord/blob/2de2887476d8973d44dc0757f579b5382e14d5b3/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java#L104) and [this](https://github.com/walmartlabs/concord/blob/master/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java#L150) - I've already merged it, but it's more or less the same amount of "leakage" :-)

I'll ponder on a better way.

amithkb: Ok.. how about having another class SsoUserInfoProvider implementation of UserInfoProvider to keep abstraction?

ibodrov: Yes, sounds like a good idea.

amithkb: refactored

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java
```diff
@@ -230,7 +219,18 @@ public void disable(UUID userId) {
                 .log();
     }
 
-    private UserInfoProvider assertProvider(UserType type) {
+    private UserInfoProvider assertProvider(String username, UserType type) {
+        /* Override SSO type if current user loggedIn via SSO */
+        if (!type.equals(UserType.SSO)){
```
ibodrov, line 224: Why do we need this? This is some strange logic to ask for a `UserInfoProvider` of the specified type and then return something else.

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java
```diff
@@ -230,7 +219,18 @@ public void disable(UUID userId) {
...
+            UserPrincipal u = UserPrincipal.getCurrent();
+            if (u != null && u.getRealm().equals(SSO_REALM_NAME) && u.getUsername().equalsIgnoreCase(username)){
```
ibodrov, line 226: What about the domain?

ibodrov (changes requested):

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java
amithkb: This override is needed because to address getInfo from [here](https://github.com/walmartlabs/concord/blob/5dd4403ca4f9e65251d7c67bc5e0d09d2e94048c/server/impl/src/main/java/com/walmartlabs/concord/server/process/pipelines/processors/UserInfoProcessor.java#L52) and [here](https://github.com/walmartlabs/concord/blob/5dd4403ca4f9e65251d7c67bc5e0d09d2e94048c/server/impl/src/main/java/com/walmartlabs/concord/server/policy/PolicyManager.java#L133) otherwise will get following error although process is trigger via UI after SSO login

image

Now, I have refactored this logic to separate method #assertUserType

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserManager.java
amithkb: considered domain

amithkb:
@concord-ci test it

ibodrov:
@concord-ci test it

ibodrov (approved):
LGTM

PR #353, opened 2021-03-10
git-client: Disable git auto-maintenance

benbroadaway:
Cached repos are never pushed back to their origin. Disable auto-maintenance since it will just keep auto-maintenance-ing.

`git maintenance` was introduced in `2.29.0` and auto-maintenance is enabled by default. It can be controlled by `git config maintenance.auto` and the `fetch` parameter `--[no-]auto-maintenance`.


brig (dismissed):

benbroadaway:
Not quite ready, centos repos are still serving up `2.27.0`

brig (approved):

PR #365, opened 2021-03-16
concord-server: Set initiator for onCancel and onFailure handlers

benbroadaway:


ibodrov:
What about `onTimeout`?

brig:
I think we should start fork (handler) with `ProcessSecurityContext.runAs(initiator, ...)`

something like we use in github events: https://github.com/walmartlabs/concord/blob/master/server/impl/src/main/java/com/walmartlabs/concord/server/events/TriggerProcessExecutor.java#L213

brig:
so just:
```
server/impl/src/main/java/com/walmartlabs/concord/server/process/queue/ProcessQueueWatchdog.java:

ProcessHandlersWorker.process:
-                processManager.startFork(payload);
+                processSecurityContext.runAs(parent.initiatorId, () -> processManager.startFork(payload));

```

benbroadaway:
@brig Yep, that's more comprehensive. Especially if some other handler is dreamt up. No need for extra fiddling.

brig (approved):

ibodrov (approved):

PR #376, opened 2021-03-25
pfed-sso-fix: add LdapPrincipal to security

amithkb:
* Absence of ldap principal breaks [restricted forms](http://concord.walmart.com/docs/getting-started/forms.html#restricting-forms) functionality
* Absence of ldap values makes `initiator` Object partial filled when process triggered by authenticating via sso

-- Fix for both


server/impl/src/main/java/com/walmartlabs/concord/server/process/form/FormAccessManager.java
```diff
@@ -129,6 +130,13 @@ private static boolean matchesLdapGroup(String pattern, Set<String> userLdapGrou
         }
 
         return userLdapGroups.stream()
-                .anyMatch(g -> Pattern.compile(pattern, Pattern.CASE_INSENSITIVE).matcher(g).matches());
+                .anyMatch(g -> Pattern.compile(normalizeGroup(pattern), Pattern.CASE_INSENSITIVE).matcher(normalizeGroup(g)).matches());
+    }
+
+    private static String normalizeGroup(String group){
```
ibodrov, line 136: I'm not so sure about this. IMO, it adds a unexpected layer of complexity: [the example in the docs](https://github.com/walmartlabs/concord-website/blame/master/docs/getting-started/forms.md#L496) assumes that the whole pattern is matched. You specify the pattern and it is matched "as-is" with the group name. But with the introduction of this change, the logic is slightly less clear - suddenly, we look inside the pattern string and decide on our own what bits of it we can ignore.

Perhaps, in practice, it won't be that big of a problem - I don't want tell our users to use `(CN=)?whatever.*` patterns either. Any thoughts? @amithkb @mtkunkel @benbroadaway 

Also, maybe don't add a JAR if it can be replaced with three lines of code? The fewer dependencies the better, IMO.

amithkb: pingfed gives groups directly like `[group1, group2...]` instead of pattern `[CN=group1,OU=..]` but however users have their flows according to doc mentioned above. Hence the change of extracting the need i,e `groupName` suitable for both cases; like with this users can provide only the group name or pattern also in `runAs` parameter. So any better way to handle this?

Will replace the jar. Noted :)

benbroadaway: Pros vs cons for pulling the groups from the `user_ldap_groups` table? Assumes the ldap group synchronizer task runs.

amithkb: I think `user_ldap_groups` may not be realtime as it is async update run at lesser frequency. AFAIK

updated the logic for `full pattern match` and handle `only group names` case. Is it good? @ibodrov 

removed jar

ibodrov (approved):

PR #384, opened 2021-04-01
ansible: limit inventory size with verbose logging

benbroadaway:
~Throws an exception~ Disables verbose log output if verbose logging is enabled and either:
1. The inventory size is too large or
2. The playbook contains too much work or
3. The playbook dynamically adds too much work (e.g. via `include_tasks`)

This limits excessive logging in large production deployments. Seems teams like to turn it on to debug something then never get around to disabling it.


benbroadaway:
Some details to hammer out:
- What should the max size be? Currently hard-coded to `50`
- Should it be configurable? May allow jerks to force their way through the check.

plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/AnsibleInventory.java
```diff
@@ -25,23 +25,94 @@
...
+                .directory(workDir.toFile());
+
+        final ExecutorService executor = Executors.newCachedThreadPool();
+
+        try {
+            Process p = pb.start();
+
+            Future<Integer> out = executor.submit(() -> {
+                try (InputStream is = p.getInputStream()) {
+                    // parse ansible-inventory JSON output
```
mtkunkel, line 80: Should `is` just be passed to ansible-playbook since it's already merged into json, to avoid subsequent execution in ansible-playbook? Some of the example inventory [scripts](https://github.com/ansible-collections/community.general/tree/main/scripts/inventory) and [plugins](https://docs.ansible.com/ansible/latest/plugins/inventory.html#inventory-plugins) take a while depending on the data source.

ibodrov:
I think this feature should be optional, but we need a way to "force enable" it for certain environments.
If we can't "force enable" it with policies directly (I need to refresh my memory) then, perhaps, we can enable the feature via the default parameters and then limit it to "enabled" using the policy.
wdyt @brig

brig:
should ansible task throw error if I have huge inventory but using only few hosts from some host group?

brig:
> I think this feature should be optional, but we need a way to "force enable" it for certain environments.
> If we can't "force enable" it with policies directly (I need to refresh my memory) then, perhaps, we can enable the feature via the default parameters and then limit it to "enabled" using the policy.
> wdyt @brig

we do not have policy for rewriting task arguments. we can use default task variables: read this option only from default variables :)

benbroadaway:
Ok, thinking about this from a different angle...doing the enforcement in an Ansible callback module. We already tally up the unique host size and even "total work" (`hosts * play tasks` for each playbook).
https://github.com/walmartlabs/concord/blob/b48162d1d0b844759762ef5fafc496a42a65f2cf/plugins/tasks/ansible/src/main/resources/com/walmartlabs/concord/plugins/ansible/callback/concord_events.py#L245-L249

Then check that against a `processCfg` policy setting the host limit or totalWork limit, similar to 
https://github.com/walmartlabs/concord/blob/b48162d1d0b844759762ef5fafc496a42a65f2cf/plugins/tasks/ansible/src/main/resources/com/walmartlabs/concord/plugins/ansible/lib/task_policy.py#L24-L30

I believe that addresses most of the issues discussed thus far.
1. Make it optional for the concord instance and configurable -> enabled via policy
2. Handle inventories with multiple groups, but limited via `--limit` -> enforced after limit is applied

benbroadaway:
@concord-ci test it

benbroadaway:
@ibodrov @mtkunkel Ok, ready for some further opinions. I changed the functionality to just disable the verbose setting rather than killing the entire playbook--less disruption to existing workflows and gets the setting to where we want it anyway.

One issue with the original approach of maxWork is that doesn't count dynamically added tasks from `import_tasks`. So I added a second check in `_on_task_start` to keep track of the ongoing work and if it turns out the playbook snuck by the initial check then we can disable verbose output as the limit it hit at runtime.

ibodrov:
@benbroadaway I've added you to the list of users allowed to do `@concord-ci test it`. It should work for you now.

ibodrov (approved):

ibodrov:
@concord-ci test it

PR #395, opened 2021-04-15
concord-server: Only pass enabled repos to refresh task

benbroadaway:
Fixes the repo search pattern. Was not including ssh urls.

In the repository refresh workflow, we previously filtered the trigger on enabled repos. But if there's a mix of enabled and disabled repos all of them would get processed.

### DIFF
--- a/server/impl/src/main/resources/com/walmartlabs/concord/server/org/triggers/concord.yml
+++ b/server/impl/src/main/resources/com/walmartlabs/concord/server/org/triggers/concord.yml
@@ -4,9 +4,14 @@ configuration:
 flows:
   onChange:
   - log: "Refreshing repository... ${event}"
-  - task: repositoryRefresh
-    in:
-      repositoryInfo: ${event.repositoryInfo.stream().filter(r -> r.enabled).toList()}
+  - name: "Filter to enabled repositories"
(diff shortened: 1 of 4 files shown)


ibodrov (dismissed):

server/impl/src/main/resources/com/walmartlabs/concord/server/org/triggers/concord.yml
```diff
@@ -6,7 +6,7 @@ flows:
   - log: "Refreshing repository... ${event}"
   - task: repositoryRefresh
     in:
-      repositoryInfo: ${event.repositoryInfo}
+      repositoryInfo: ${event.repositoryInfo.stream().filter(r -> r.enabled).toList()}
```
brig, line 9: https://github.com/walmartlabs/concord/blob/master/plugins/tasks/concord/src/main/java/com/walmartlabs/concord/client/RepositoryRefreshTaskParams.java#L45

throws exception for empty repo list :)

benbroadaway: Ooooh good catch

brig (approved):

brig (approved):

PR #410, opened 2021-05-06
db: initialize with dynamic default agent token

benbroadaway:
- Initialize DB without a hard-coded default agent API token:  New DBs will now initialize with a random or config-provided default API token. Existing DBs that update to the next version will not be cleaned up--so if anyone is actually still using the old hard-coded tokens they need to manually rotate the agent API token in their agent config.
- Modify dev scripts (docker-compose, vagrant) to handle agent API token
- Remove `concordRunner` user and tokens: It's not used


benbroadaway:
Requires some changes to testcontainers-concord https://github.com/concord-workflow/testcontainers-concord/pull/23

ibodrov (approved):

PR #412, opened 2021-05-10
concord-server: add DNS Service Record feature ldap (dynamic loading of LDAP server address)

amithkb:
feature to discover ldap AD servers dynamically according to [this](https://ldap.com/dns-srv-records-for-ldap/) doc


mtkunkel:
> * Initialize with a fresh SRV list
> * Make LDAP calls with a "good" server, if one fails...
> * Get alternate LDAP server from remaining hosts in original SRV list
> * If no more "good" servers, refresh SRV list

Potentially get the list and keep it cached (to avoid doing multiple dns srv lookups).  Create a fresh, disposable LDAPContext with each LDAP call using the cached server list?  Assumes disposable contexts are don't cause other performance issues...

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/ConcordLdapContextFactory.java
```diff
@@ -23,26 +23,118 @@
...
+            return getLdapContext(username, password);
+        }
+    }
+
+    @Override
+    public LdapContext getLdapContext(Object principal, Object credentials) throws NamingException {
+        try {
+            return this.delegate.getLdapContext(principal, credentials);
+        } catch (CommunicationException e) {
+            recursionLimiter();
```
ibodrov, line 116: This is concerning. In which situations this call might be recursive?

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/Utils.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =====
+ */
+
+public class Utils {
+    public static final int maxLevel = 20;
+
+    public static void recursionLimiter() {
+        if (maxLevel == 0)
```
ibodrov, line 27: I think this condition is never true?

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/ConcordLdapContextFactory.java
```diff
@@ -53,26 +145,69 @@ protected LdapContext createLdapContext(Hashtable env) throws NamingException {
...
+            }
+        }
+        return servers;
     }
 
-    @Override
-    public LdapContext getSystemLdapContext() throws NamingException {
-        return delegate.getSystemLdapContext();
+    private static String removeLastCharIfDot(String s) {
+        if (s == null || s.length() == 0)
```
ibodrov, line 176: Needs curly braces.

ibodrov (changes requested):

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/ConcordLdapContextFactory.java
amithkb: no good servers(may be decommissioned/removed off) in the list then call gets recursive to check for each server

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/Utils.java
amithkb: yes. just added logic if we don't want recursion at all then setting `maxLevel = 0` will work. not necessary?

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/ConcordLdapContextFactory.java
ibodrov: Why it cannot be a loop? Sifting through stacktraces seems a bit unconventional.

amithkb: okay. was trying to reduce code duplication. [Replaced](https://github.com/walmartlabs/concord/pull/412/commits/916fae77478346b7620149bce8f318b39c53e014) with loop approach to achieve depth limit

ibodrov (approved):

ibodrov:
Nitpicking: the implementation is still recursive. I'm not sure how important it is on practice, but it can be a potential future improvement. E.g. have the iterator/failure handing core in one place by replacing `getSystemLdapContext` with a loop over `ldapUrlIterator`, advancing the iterator on every failure instead of calling `getLdapContext` recursively. And maybe splitting `LdapContextFactory` implementation into the original one (supports only a single LDAP URL) and the new SRV-based one.
Regardless, it's probably not very important right now.

PR #433, opened 2021-06-23
console2: fix json store query editor

brig:
after version up of monaco-editor


amithkb (approved):

benbroadaway:
Does this affect the Project configuration editor too? Seems that's broken as well.

brig:
> Does this affect the Project configuration editor too? Seems that's broken as well.

yep

benbroadaway (approved):

PR #442, opened 2021-07-14
concord-server: add user ldapgroup sync api

amithkb:
API to sync ldap groups for a user

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/UserLdapGroup.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/UserLdapGroup.java
@@ -1,0 +1,89 @@
+package com.walmartlabs.concord.server.security.ldap;
+
+import com.walmartlabs.concord.server.GenericOperationResult;
+import com.walmartlabs.concord.server.OperationResult;
+import com.walmartlabs.concord.server.sdk.ConcordApplicationException;
+import com.walmartlabs.concord.server.security.Roles;
+import com.walmartlabs.concord.server.user.UserInfoProvider;
+import com.walmartlabs.concord.server.user.UserManager;
+import com.walmartlabs.concord.server.user.UserType;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.Authorization;
+import org.apache.shiro.authz.UnauthorizedException;
+import org.sonatype.siesta.Resource;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.Collections;
+import java.util.Set;
+import java.util.UUID;
+
+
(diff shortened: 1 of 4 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/UserLdapGroup.java
```diff
@@ -0,0 +1,95 @@
...
+        UserInfoProvider.UserInfo info = ldapUserInfoProvider.getInfo(id, username, domain);
+        if (info == null) {
+            throw new ConcordApplicationException("User '" + username + "' with domain '" + domain + "' not found in LDAP", Response.Status.BAD_REQUEST);
+        }
+
+        try {
+            Set<String> groups = ldapManager.getGroups(username, domain);
+            if (groups == null) {
+                ldapGroupsDao.update(id, Collections.emptySet());
+                ldapGroupsDao.updateLastSyncTimestamp(id);
```
ibodrov, line 79: I think `ldapGroupsDao.update` already calls `updateLastSyncTimestamp` by itself? Perhaps we should make `updateLastSyncTimestamp` private and make sure the rest of the code uses only the `update` method?

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/UserLdapGroup.java
```diff
@@ -0,0 +1,95 @@
...
+     * Sync Ldap groups for a ldap user
+     *
+     * @param username username
+     * @param domain   domain
+     * @return GenericOperationResult result
+     */
+    @POST
+    @ApiOperation("Sync ldap groups for a user")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/user/{username}/domain/{domain}/sync")
```
ibodrov, line 60: Not sure why would a domain be a sub-resource of a user?
Perhaps it should be a
```
POST /api/v1/userldapgroup/sync
{
  "username": "...",
  "domain": "..."
}
```
request instead?

amithkb: updated

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/UserLdapGroup.java
amithkb: updated

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/UserLdapGroup.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param req user's data
+     * @return GenericOperationResult result
+     */
+    @POST
+    @ApiOperation("Sync ldap groups for a user")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/sync")
+    @Validate
+    public GenericOperationResult sync(@ApiParam @Valid CreateUserRequest req) {
```
ibodrov, line 85: I think it would be better to use a new type here instead of re-using `CreateUserRequest`. There are fields in the `CreateUserRequest` class that are not relevant for this method.

amithkb: sounds good. updated

ibodrov (approved):

PR #446, opened 2021-07-16
concord-server: fix for duplicates in status history

amithkb:
fix for issue
image


ibodrov:
We need to fix the duplicate event insertion, not the query. @brig and I are looking into it.

ibodrov:
Closing in favor of #450

PR #447, opened 2021-07-19
runtime-v2: allow expressions as input parameters for task/flow/script

brig:


runtime/v2/model/src/test/resources/schema/concord.yml
```diff
@@ -47,8 +47,8 @@ flows:
     - form: "formName"
     - throw: ${error}
     - call: "default2"
-      in:
-        k: "v"
+#      in:
+#        k: "v"
```
benbroadaway, line 51: Is this intended be be left commented out?

brig: it's commented because `ConcordJsonSchemaGeneratorTest` failed. do not know why, maybe bug in json validator. do not want to waste time on this, so just commented this lines :)

benbroadaway (approved):

ibodrov (approved):

PR #448, opened 2021-07-20
examples: more ansible retry examples

benbroadaway:
Two more example flows for retrying with a limit file:
1. Automatically after suspending for some amount of time
2. On-demand after a form


examples/ansible_retry/concord.yml
```diff
@@ -28,3 +35,100 @@ flows:
...
+                  - "127.0.0.3"
+                vars:
+                  ansible_connection: "local"
+            extraVars:
+              makeItFail: "${makeItFail}"
+      error:
+        - log: "Ansible error: ${lastError.cause}"
+
+        # Save retry file as an attachment
+        - call: copyRetryFile
```
ibodrov, line 60: I think `saveRetryFile` can be used instead? https://github.com/walmartlabs/concord/blob/master/plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/TaskParams.java#L108

ibodrov (approved):

examples/ansible_retry/concord.yml
benbroadaway: Sure, makes sense 👍

ibodrov: Sorry, I meant the `saveRetry` parameter of the task. It saves the retry file as a process attachment, so you don't have to copy it manually.

benbroadaway: Hmmm I think we have a bug with that parameter. It's copying`<playbook>.yml` to `_attachments/ansible.retry`, but it _should_ be copying `<playbook>.retry` to `_attachments/<playbook>.retry`

benbroadaway: Yeah https://github.com/walmartlabs/concord/blob/cc59284f6840f236305342296b2ce38dfae21578/plugins/tasks/ansible/src/main/java/com/walmartlabs/concord/plugins/ansible/AnsibleTask.java#L230 should be
```java
String retryFile = getNameWithoutExtension(playbookName) + ".retry";
```

benbroadaway: https://github.com/walmartlabs/concord/pull/451 PR for fix. Will update this example afterwards

ibodrov (approved):

PR #451, opened 2021-07-27
ansible: fix retry file persistance

benbroadaway:
- Fix bug where ansible plugin copied the playbook instead of the `.retry` file to `_attachments/`.
- Use the actual retry file name rather than hard-coded retry file name. Useful if multiple playbooks are run and need to be retried.


ibodrov:
@concord-ci test it

ibodrov (approved):

PR #453, opened 2021-08-01
runtime-v2: copy variables in try block into parent frame

ibodrov:
Variables in `try` blocks created before the error must be visible in the outer block (the block that contains the `try`
block).

Copy inner frame (the try block's body) variables into the outer block when the thread is unwinding.

### DIFF
--- a/runtime/v2/runner/src/test/resources/com/walmartlabs/concord/runtime/v2/runner/errorBlockScoping/concord.yml
+++ b/runtime/v2/runner/src/test/resources/com/walmartlabs/concord/runtime/v2/runner/errorBlockScoping/concord.yml
@@ -1,0 +1,23 @@
+configuration:
+  runtime: "concord-v2"
+
+flows:
+  default:
+    - call: anotherFlow
+      out:
+        - myVar
+    - log: "Hello, ${myVar}!"
+
+  anotherFlow:
+    - try:
+        - call: createMyVar
+          out:
+            - myVar
+        - throw: kapow! kamblamo!
+      error:
+        - log: "caught an error"
+    - log: "myVar should exist: ${myVar}"
+
+  createMyVar:
+    - set:
+        myVar: 'world'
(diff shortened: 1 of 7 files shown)


runtime/v2/runner/src/test/resources/com/walmartlabs/concord/runtime/v2/runner/errorBlockScoping/concord.yml
```diff
@@ -0,0 +1,23 @@
...
+
+  anotherFlow:
+    - try:
+        - call: createMyVar
+          out:
+            - myVar
+        - throw: kapow! kamblamo!
+      error:
+        - log: "caught an error"
+    - log: "myVar should exist: ${myVar}"
```
benbroadaway, line 19: This should still work even when the `throw` call is omitted

benbroadaway (changes requested):

runtime/v2/runner/src/test/resources/com/walmartlabs/concord/runtime/v2/runner/errorBlockScoping/concord.yml
ibodrov: Right, I initially though the issue was limited to the `error` blocks. I'll update the PR tomorrow or on Wednesday.

benbroadaway: Another similar scenario. Putting an `error` block on a flow call messes up the variable scope too.
```yaml
flows:
  default:
    - call: anotherFlow
      out:
        - myVar
    - log: "Hello, ${myVar}!"

  anotherFlow:
    - call: createMyVar
      out:
        - myVar
      error:
        - log: "caught an error"
    - log: "myVar should exist: ${myVar}"  # doesn't work

  createMyVar:
    - set:
        myVar: 'world'
```

brig:
fixed in this PR: https://github.com/walmartlabs/concord/pull/496

PR #457, opened 2021-08-09
concord-server: load API keys from a local file

ibodrov:
Provides a way to load user API tokens from a local file
(e.g. from a secret file mounted into the server container).

Unlike `defaultAdminToken` or `defaultAgentToken` parameters in concord-server.conf,
this mechanism allows much easier token rotation.

Usage:
```
# concord-server.conf
concord-server {
    apiKey {
        loadFrom = "/path/to/api_keys.json"
    }
}
```

or `export CONCORD_API_KEYS_FILE=/path/to/api_keys.json` before starting the server.

The file format:
```json
[
    {
        "username": "admin",
        "keyName": "default",
        "value": "...base64...",
        "expiredAt": "2022-01-02T03:04:05.678-04"
    },
    {
        "username": "admin",
        "keyName": "second_one",
        "value": "...base64..."
    }
]
```

The `expiredAt` parameter is optional.
The `keyName` and `value` must be globally unique.

Users from multiple realms (LDAP, LOCAL, etc) are not supported, e.g. there should be only one user with the username `admin`.

Additionally, now it is possible to skip the default token generation for admin and/or agent:
```
# concord-server.conf
concord-server {
    db {
        changeLogParameters {
            skipAdminTokenGeneration = "true"
            skipAgentTokenGeneration = "true"
        }
    }
}
```


ibodrov:
Actually, thinking about it, I can extend this mechanism to be able to define users in addition to user API keys.

PR #471, opened 2021-09-02
server: do not create http sessions for api-key auth

brig:


benbroadaway:
Looks like the "login with API token" option makes this a little trickier.

benbroadaway (approved):

PR #475, opened 2021-09-17
agent: do not resend logs when server response with 400-500 error codes (e.g. 401)

brig:


amithkb (approved):

agent/src/main/java/com/walmartlabs/concord/agent/logging/RemoteLogAppender.java
```diff
@@ -73,7 +73,8 @@ public boolean appendLog(UUID instanceId, long segmentId, byte[] ab) {
             return true;
         } catch (ApiException e) {
             log.warn("appendLog ['{}'] -> error: {}", instanceId, e.getMessage());
-            return false;
+
+            return e.getCode() >= 400 && e.getCode() < 500;
```
benbroadaway, line 77: This does get it to stop eventually, but should we consider doing something more drastic like throw a runtime exception or interrupt the thread? When testing locally, my agent fires off an extra 100+ calls before it finally gets cleaned up.

And am I wrong...shouldn't we add this to the other methods here [`appendLog`](https://github.com/walmartlabs/concord/blob/3192f97fefca9371884afeeb176f6baebc0acbf9/agent/src/main/java/com/walmartlabs/concord/agent/logging/RemoteLogAppender.java#L58-L61) and [`updateSegment`](https://github.com/walmartlabs/concord/blob/3192f97fefca9371884afeeb176f6baebc0acbf9/agent/src/main/java/com/walmartlabs/concord/agent/logging/RemoteLogAppender.java#L92-L94)?

brig: I thought about it. We have `CombinedLogAppender` with multiple appenders. If one appender failed should other appender try to log? I think that we need to reorganize the work of the logs in the agent, will do it in another PR.

return value from `updateSegment` is never used and `appendLog` has no return value :)

benbroadaway (approved):

PR #484, opened 2021-11-10
runtime-v2: support for task `ignoreErrors` mode

brig:
+ allows task to return variables even if the task failed


ibodrov:
@concord-ci test it

ibodrov (approved):

PR #489, opened 2021-11-16
docker: more input params to handle logging

### LINKED ISSUE #485: docker-task: broken runtime-v2 stderr capture
`stderr` output capturing is broken in DockerTaskV2. The main issue is that `stdout` and `stderr` are combined by default`

https://github.com/walmartlabs/concord/blob/228b21256f5156bb0eb2e76a3143ce2318c198b2/plugins/tasks/docker/src/main/java/com/walmartlabs/concord/plugins/docker/DockerTaskV2.java#L60

Then, when using `logOutput: false`, the output isn't combined...but it also isn't captured at all

https://github.com/walmartlabs/concord/blob/228b21256f5156bb0eb2e76a3143ce2318c198b2/plugins/tasks/docker/src/main/java/com/walmartlabs/concord/plugins/docker/DockerTaskV2.java#L80-L85

brig:
new input variables:
```
redirectErrorStream - optional, boolean. If `true` docker std error stream redirects to std out; 
logOut - optional, boolean. If `true` Concord logs docker std out to process logs;
logErr - optional, boolean. If `true` Concord logs docker std error to process logs;
saveOut - optional, boolean. If `true` docker task returns std out as task result;
saveErr - optional, boolean. If `true` docker task returns std err as task result;
```

### DIFF
--- a/plugins/tasks/docker/src/main/java/com/walmartlabs/concord/plugins/docker/DockerTaskV2.java
+++ b/plugins/tasks/docker/src/main/java/com/walmartlabs/concord/plugins/docker/DockerTaskV2.java
@@ -40,7 +40,11 @@ public class DockerTaskV2 implements Task {
     private static final Logger log = LoggerFactory.getLogger(DockerTask.class);
     private static final Logger processLog = LoggerFactory.getLogger("processLog");
 
-    private static final String LOG_OUTPUT_KEY = "logOutput";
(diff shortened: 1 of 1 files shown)


ibodrov (approved):

brig:
docs: https://github.com/walmartlabs/concord-website/pull/101

PR #496, opened 2021-11-24
runtime-v2: `error` block on a flow call messes up the out variables

brig:


runtime/v2/runner/src/test/resources/com/walmartlabs/concord/runtime/v2/runner/callWithErrorBlock/concord.yml
```diff
@@ -0,0 +1,13 @@
+flows:
+  default:
+    - call: createMyVar
+      out:
+        - myVar
+      error:
+        - log: "caught an error"
+    - log: "myVar should exist: ${myVar}"  # doesn't work
```
ibodrov, line 8: The comment is no longer relevant, I think?

ibodrov (approved):

runtime/v2/runner/src/test/resources/com/walmartlabs/concord/runtime/v2/runner/callWithErrorBlock/concord.yml
brig: ouch. fixed.

ibodrov (approved):

PR #502, opened 2021-12-02
concord-server: move com.walmartlabs.concord.server.ansible.* into ansible plugin

brig:


ibodrov (dismissed):

ibodrov:
We need a decision of the future structure of the server SDK/SPI wrt to the queue processors.
We'd like to introduce a concept of "phases" (like what Maven does) into the server's process pipelines (e.g. EnqueueProcessPipeline).

ibodrov (approved):

PR #517, opened 2021-12-17
runtime-v2: disable warning for graalvm, fix logging

brig:
- disable graalvm warining:
```
WARNING: The polyglot context is using an implementation that does not support runtime compilation...
```


brig:
@concord-ci test it

ibodrov (approved):

PR #540, opened 2022-01-13
runtime-v2: script out parameter

brig:


benbroadaway (approved):

benbroadaway:
There's a very tiny risk of an existing flow/script that relies on injected script variables that may break
```yaml
    - set:
        result: "result from another call"
    - script: groovy
      body: |
         // result is ScriptResult, not the flow variable
        println("result => " + result)
```

The easy workaround is to get it from the `execution` variable:
```yaml
    - set:
        result: "result from another call"
    - script: groovy
      body: |
        def theResult = execution.variables.getString('result')
        println("result => " + theResult)
```

PR #573, opened 2022-02-24
runtime-v1/v2: event configuration option to evaluate checkpoint names

benbroadaway:
This is a little trickier than it seems at first. The process process event recorder doesn't get direct access to the checkpoint name when it's created--it gets the value from the YAML file which is typically a hard-coded string, but may be an expression. The PR exposes an option in the runner config to enable evaluation of expressions in checkpoint names. It's off by default.

In runtime-v2, it's limited to only expressions using variables. If a checkpoint name's expression is uses a task call, e.g. `- checkpoint: ${resource.asJson('myList.json')[0]}`, then the expression will evaluate to `null` and fallback to the default from-the-yaml value.


benbroadaway:
Closing in favor of https://github.com/walmartlabs/concord/pull/581

PR #581, opened 2022-03-07
event correlationId into process checkpoint

brig:


benbroadaway (dismissed):

runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/ConcordJsonSchemaGenerator.java
```diff
@@ -68,7 +68,7 @@ public static JsonNode generate() {
         switchStep.set("additionalProperties", switchDefault);
 
         // remove invalid required primitive attributes
-        removeRequired(path(jsonSchema, "definitions/ProcessDefinitionConfiguration"), "debug");
+        removeRequired(path(jsonSchema, "definitions/ProcessDefinitionConfiguration"), "debug", "parallelLoopParallelism");
```
benbroadaway, line 71: this is just some extra cleanup right?

brig: yep. current version of JsonSchemaGenerator lib marks booleans as required fields.

ibodrov (approved):

PR #597, opened 2022-03-29
http-task: proxy auth

brig:
```
  - task: http
    in:
      method: GET
      url: "http://httpbin.org/get"
      proxy: "http://proxy:8080"
      proxyAuth:
        user: "user"
        password: "pass"
```

### DIFF
--- a/plugins/tasks/http/src/main/java/com/walmartlabs/concord/plugins/http/Configuration.java
+++ b/plugins/tasks/http/src/main/java/com/walmartlabs/concord/plugins/http/Configuration.java
@@ -28,6 +28,7 @@ import com.walmartlabs.concord.sdk.MapUtils;
 import org.apache.http.client.utils.URIBuilder;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -55,6 +56,8 @@ public class Configuration {
     private final int requestTimeout;
     private final boolean ignoreErrors;
     private final String proxy;
+    private final String proxyUser;
+    private final String proxyPassword;
     private final boolean debug;
     private boolean followRedirects;
     private final String keyStorePath;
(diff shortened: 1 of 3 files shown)


plugins/tasks/http/src/main/java/com/walmartlabs/concord/plugins/http/Configuration.java
```diff
@@ -55,6 +56,8 @@ public class Configuration {
     private final int requestTimeout;
     private final boolean ignoreErrors;
     private final String proxy;
+    private final String proxyUser;
+    private final String proxyPassword;
```
ibodrov, line 60: Use `chat[]`? A good practice to prevent string interning.

ibodrov (changes requested):

plugins/tasks/http/src/main/java/com/walmartlabs/concord/plugins/http/Configuration.java
brig: fixed

benbroadaway: I'm a little confused on the issue here. Why should we try to prevent interning on just these two constant strings rather than all the others (e.g. `proxy`, `keyStorePath`, `keyStorePassword`)?

benbroadaway: ...is it just a general rule of "don't intern credentials"

brig: saw keyStorePassword and do not wanna change it in this PR...

there are a lot of answers why char array is better:
https://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords

ibodrov (approved):

PR #608, opened 2022-06-15
server: fix out vars processing and restrictions

benbroadaway:
The existing tests were broken in the bad way (passed but for the wrong reasons).

The out-vars restrictions haven't been applying to `out` vars specified in a `request` file which is how the concord plugin passes them to child processes.


ibodrov (approved):
LGTM

The only (very minor) downside I can think of is the double parse of _main.json and the request JSON files.

ibodrov:
Closing in favor of #609

PR #609, opened 2022-06-16
server: fix out vars processing and restrictions (alternative)

benbroadaway:
Alternative to https://github.com/walmartlabs/concord/pull/608

Pros and cons...This approach doesn't double-parse the project. Instead it double-checks out variables.
1. Checks for out variables given in direct request parameters (e.g. `curl https://.../api/v1/process?out=myVarA,myVarB`) in the `NewProcessPipeline`
2. _After_ the project is parsed in `EnqueueProcessPipeline`, all the `out` variables are consolidated and the check is made again for out variables being added in a project that doesn't allow it.

The main difference is that the process entry gets created here and fails immediately if it fails on the second check. In https://github.com/walmartlabs/concord/pull/608, the process entry never gets created at all.

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/process/pipelines/processors/ConfigurationProcessor.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/process/pipelines/processors/ConfigurationProcessor.java
@@ -42,10 +42,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
 
 /**
  * Responsible for preparing the process' {@code configuration} object.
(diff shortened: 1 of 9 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/process/pipelines/processors/ConfigurationProcessor.java
```diff
@@ -90,6 +87,10 @@ public Payload process(Chain chain, Payload payload) {
         List<String> activeProfiles = ProcessConfigurationUtils.getActiveProfiles(payloadCfg, attachedCfg, workspaceCfg, projectCfg, orgCfg);
         payload = payload.putHeader(Payload.ACTIVE_PROFILES, activeProfiles);
 
+        // consolidate out variables
+        Set<String> outVars = new HashSet<>(ProcessConfigurationUtils.getOutVars(payloadCfg, attachedCfg, workspaceCfg));
```
ibodrov, line 91: I might be reading it wrong, but it looks like `getOutVars(...)` returns the first occurrence of `out` rather than a merged set of all `out` values?

ibodrov (approved):

PR #610, opened 2022-06-20
ansible:  sort options on ansible host stats

amithkb:
image

### DIFF
--- a/server/plugins/ansible/impl/src/main/java/com/walmartlabs/concord/server/plugins/ansible/ProcessAnsibleResource.java
+++ b/server/plugins/ansible/impl/src/main/java/com/walmartlabs/concord/server/plugins/ansible/ProcessAnsibleResource.java
@@ -562,7 +562,8 @@ public class ProcessAnsibleResource implements Resource {
     public enum SortField {
         host,
         duration,
-        status
+        status,
+        host_group
     }
     
     public enum SortBy {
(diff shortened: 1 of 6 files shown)


server/plugins/ansible/impl/src/main/java/com/walmartlabs/concord/server/plugins/ansible/ProcessAnsibleResource.java
```diff
@@ -327,9 +329,18 @@ public List<AnsibleHostEntry> list(UUID instanceId, OffsetDateTime instanceCreat
                 if (playbookId != null) {
                     q.and(a.PLAYBOOK_ID.eq(playbookId));
                 }
-
-                return q.orderBy(a.HOST)
-                        .limit(limit)
+                
+                if (sortField != null) {
+                    if (sortBy != null && sortBy.equals(SortBy.DESC)) {
+                        q.orderBy(Objects.requireNonNull(a.field(sortField.name())).desc());
```
ibodrov, line 335: Maybe add a message to `requireNonNull`, so if it fails it will be more apparent on which field it failed.
E.g.
```java
Field orderBy = requireNonNull(a.field(sortField.name()), "Invalid sort field: " + sortField.name()).desc();
q.orderBy(orderBy)
``` 

I think it is even possible to do such validation outside of the transaction.

server/plugins/ansible/impl/src/main/java/com/walmartlabs/concord/server/plugins/ansible/ProcessAnsibleResource.java
```diff
@@ -547,4 +558,16 @@ public enum TaskType {
         SETUP,
         HANDLER
     }
+    
+    public enum SortField {
```
ibodrov, line 562: Maybe make enum values uppercase for consistency's sake and convert to lower later?

ibodrov (commented):
LGTM, just a couple of minor comments.

ibodrov (dismissed):

ibodrov (approved):

mtkunkel (approved):

PR #623, opened 2022-07-12
server: runtime-v2 custom forms file field fix

benbroadaway:
File upload in runtime-v2 custom forms is broken. The submitted data is being converted twice. the first conversion writes the uploaded file to a temp directory for upload to process_state (that's good). But the second run ends up overwriting the file with the filename as the contents.

Solution is to only convert once, which is what non-custom form service and runtime-v1 do.


brig (approved):

ibodrov (approved):
👍

PR #634, opened 2022-07-23
concord-console: adjust polling frequency based on client activity

benbroadaway:
Introduces polling intervals based on user (in)activity and window visibility.

1. 5s - Window is visible and user is active
2. 60s - Window is visible and user is idle (no activity for 10 minutes) OR window is not visible


dankle (approved):

benbroadaway:
polling

ibodrov (approved):

PR #635, opened 2022-07-24
chore: use mvnw wrapper jars

ajayk:


ajayk:
@ibodrov : 
@jvanzyl archived the takari wrapper https://github.com/takari/maven-wrapper and the path is to use the https://github.com/apache/maven-wrapper 

@jvanzyl any recommendations here ?

jvanzyl:
I usually just use the JAR

PR #637, opened 2022-07-27
Feature to tag multiple projects to secret

hemant0308:
https://jira.walmart.com/browse/STRDTORC-3725


hemant0308:
Attaching video demo of the ui changes.

https://user-images.githubusercontent.com/13145963/181352375-40be7c96-1b62-42c4-a66c-7ae17b43d45d.mov

plugins/tasks/concord/src/main/java/com/walmartlabs/concord/client/SecretsTaskParams.java
```diff
@@ -112,13 +112,15 @@ static class CreateParams extends SecretsTaskParams {
...
+        static final String PROJECT_NAMES_KEY = "projects";
+        static final String PROJECT_IDS_KEY = "projectIds";
 
         private CreateParams(Variables variables) {
             super(variables);
         }
 
         public SecretEntry.TypeEnum secretType() {
-            String type = variables.getString(Constants.Multipart.TYPE, SecretEntry.TypeEnum.DATA.toString());
+            String type = variables.assertString(Constants.Multipart.TYPE);
```
brig, line 123: looks like breaking change. before this changes `secretType` was optional with `DATA` type as default.

server/db/src/main/resources/com/walmartlabs/concord/server/db/v1.96.0.xml
```diff
@@ -29,4 +29,32 @@
...
+    <changeSet id="1960011" author="g0h04k0walmart.com">
+        <createTable tableName="SECRET_PROJECTS">
+            <column name="SECRET_ID" type="uuid">
+                <constraints nullable="false"/>
+            </column>
+            <column name="PROJECT_ID" type="uuid">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addForeignKeyConstraint baseTableName="SECRET_PROJECTS"
```
brig, line 41: maybe add unique constraint, indexes?

server/impl/src/main/java/com/walmartlabs/concord/server/org/secret/SecretEntryV2.java
```diff
@@ -0,0 +1,156 @@
...
+import com.walmartlabs.concord.common.validation.ConcordKey;
+import com.walmartlabs.concord.server.org.EntityOwner;
+import com.walmartlabs.concord.server.org.project.ProjectEntry;
+
+import javax.validation.constraints.NotNull;
+import java.io.Serializable;
+import java.util.*;
+
+@JsonInclude(Include.NON_NULL)
+public class SecretEntryV2 implements Serializable {
```
brig, line 35: I think that we can add new fields into `SecretEntry` and mark, `project, projectName` as deprecated.
Because `SecretEntry` with project/projectName make no sense...

hemant0308: This is to not touch the old api's. If someone using the old api and not expecting these new attribute (projects), its gonna be a problem. Eventually we can deprecate the V1.

plugins/tasks/concord/src/main/java/com/walmartlabs/concord/client/SecretsTaskParams.java
hemant0308: Updated

PR #644, opened 2022-08-23
docker-images: set /bin/sh to bash from dash in debian based images

amithkb:
default sh is configured to dash in debian/ubuntu. Change to bash for compatibility


.github/workflows/ci.yml
```diff
@@ -11,6 +11,11 @@ jobs:
 
     runs-on: self-hosted
 
+    strategy:
+      fail-fast: false
+      matrix:
+        baseos: [ centos, debian ]
```
ibodrov, line 17: This is going to double the build time.
I think it should be implemented as a separate GHA flow to just to test Docker build.

ibodrov (changes requested):

ibodrov (approved):
Awesome, thanks!

PR #652, opened 2022-09-20
concord-cli: ensure absolute target dir

benbroadaway:
Concord's project loader normalizes the resource glob with an absolute path.

https://github.com/walmartlabs/concord/blob/66fc4e40af7daf4ce61fd004906eba183d7c42d1/runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/ProjectLoaderV2.java#L209

However, the CLI linter currently treats the command-line supplied `targetDir` as-is which breaks the resource filtering inside the project loader.

### DIFF
--- a/cli/src/test/java/com/walmartlabs/concord/cli/LintTest.java
+++ b/cli/src/test/java/com/walmartlabs/concord/cli/LintTest.java
@@ -4,7 +4,7 @@ package com.walmartlabs.concord.cli;
  * *****
  * Concord
  * -----
- * Copyright (C) 2017 - 2020 Walmart Inc.
+ * Copyright (C) 2017 - 2022 Walmart Inc.
  * -----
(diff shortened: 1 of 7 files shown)


cli/src/test/java/com/walmartlabs/concord/cli/LintTest.java
```diff
@@ -0,0 +1,73 @@
...
+        int exitCode = lint("lintV1");
+        assertEquals(0, exitCode);
+        assertLog(".*flows: 2.*");
+    }
+
+    @Test
+    void lintV2Test() throws Exception {
+        int exitCode = lint("lintV2");
+        assertEquals(0, exitCode);
+        assertLog(".*flows: 2.*");
```
cdaringe, line 49: ❤️

brig (approved):

PR #656, opened 2022-09-22
concord-server: termintate process wait watchdog loop on batches less than fetch limit

benbroadaway:
In a busy environment, it's possible for wait conditions to come in faster than they can be dispatched in a loop. This keeps the loop going with a small (maybe just one) element in each loop. Then that keeps processes (checked in an earlier iteration) that have been ready to resume for a while to be ignored until the `processWaits()` method gets lucky and returns nothing.


ibodrov (approved):
Nice catch!

brig (approved):

PR #658, opened 2022-09-24
concord-server: @Inject refactoring (part 1)

ibodrov:
I'm gradually replacing the usage of `@Named` autowiring with explicit bindings using `Module`.

The idea is that the top-level modules (e.g. `ConcordServer`) and server plugins are autowired by Sisu while all internal modules are wired explicitly using Guice `Module`s. This should provide better re-usability of pieces of concord-server -- e.g., for testing and creating Concord-based tools.


ibodrov:
Don't squash those commits, please.

brig (approved):

PR #691, opened 2022-12-12
Added migration task to update secrets data with new SHA algorithm

hemant0308:


server/db/src/main/java/com/walmartlabs/concord/db/migration/SecretsHashMigrationTask.java
```diff
@@ -0,0 +1,105 @@
...
+    private byte[] serverPwd;
+
+    @Override
+    synchronized public void execute(DataSource dataSource) {
+        log.info("Starting migration task for secretsHash");
+        int retryCount = 0;
+        while(retryCount <= MAX_RETRIES) {
+            try(Connection con = dataSource.getConnection()) {
+                while(true) {
+                    try(PreparedStatement psSelect = con.prepareStatement("SELECT secret_id, secret_salt, secret_data FROM secrets WHERE hash_algorithm = ? AND encrypted_by = ? LIMIT ?")){
```
brig, line 60: I think there should be a "select to update" so that no one can change the value during the update.

server/db/src/main/java/com/walmartlabs/concord/db/migration/SecretsHashMigrationTask.java
```diff
@@ -0,0 +1,105 @@
...
+                        psSelect.setString(1, HashAlgorithm.LEGACY_MD5.getName());
+                        psSelect.setString(2, SecretEncryptedByType.SERVER_KEY.toString());
+                        psSelect.setInt(3, BATCH_SIZE);
+                        ResultSet resultSet = psSelect.executeQuery();
+                        int count = 0;
+                        try(PreparedStatement psUpdated = con.prepareStatement("UPDATE secrets SET secret_data = ?, hash_algorithm = ? WHERE secret_id = ?")) {
+                            while(resultSet.next()) {
+                                String secretId = resultSet.getString(1);
+                                byte[] secretSalt = resultSet.getBytes(2);
```
brig, line 69: secret_salt is nullable in DB. I think we should check for null and use value from cfg.

server/db/src/main/java/com/walmartlabs/concord/db/migration/SecretsHashMigrationTask.java
```diff
@@ -0,0 +1,105 @@
...
+                                byte[] encryptedData = resultSet.getBytes(3);
+                                PGobject uuidObject = new PGobject();
```
brig, line 71: we can use string and explicitly cast it to uuid with `::uuid`

server/db/src/main/java/com/walmartlabs/concord/db/migration/SecretsHashMigrationTask.java
```diff
@@ -0,0 +1,105 @@
...
+        log.info("Starting migration task for secretsHash");
+        int retryCount = 0;
+        while(retryCount <= MAX_RETRIES) {
+            try(Connection con = dataSource.getConnection()) {
+                while(true) {
+                    try(PreparedStatement psSelect = con.prepareStatement("SELECT secret_id, secret_salt, secret_data FROM secrets WHERE hash_algorithm = ? AND encrypted_by = ? LIMIT ?")){
+                        psSelect.setString(1, HashAlgorithm.LEGACY_MD5.getName());
+                        psSelect.setString(2, SecretEncryptedByType.SERVER_KEY.toString());
+                        psSelect.setInt(3, BATCH_SIZE);
+                        ResultSet resultSet = psSelect.executeQuery();
```
brig, line 64: it's better to close result set when it's not needed

brig, line 64: it's better to close result set when it's not needed

brig (changes requested):

server/db/src/main/java/com/walmartlabs/concord/db/migration/SecretsHashMigrationTask.java
hemant0308: Updated

server/db/src/main/java/com/walmartlabs/concord/db/migration/SecretsHashMigrationTask.java
hemant0308: Updated

server/db/src/main/java/com/walmartlabs/concord/db/migration/SecretsHashMigrationTask.java
hemant0308: Updated

server/db/src/main/java/com/walmartlabs/concord/db/migration/SecretsHashMigrationTask.java
hemant0308: Updated

brig:
maybe convert migration task to liquibase [`CustomTaskChange`](https://docs.liquibase.com/change-types/custom-change.html) ? so we don't care about lifecycle of this task

hemant0308:
> maybe convert migration task to liquibase [`CustomTaskChange`](https://docs.liquibase.com/change-types/custom-change.html) ? so we don't care about lifecycle of this task

Converted it to customchangeTask.

server/dist/src/main/resources/concord-server.conf
```diff
@@ -112,7 +112,9 @@ concord-server {
             # requires "CREATE EXTENSION" privileges
             createExtensionAvailable = "true"
 
-            secretStoreSalt = "${secretStore.secretStoreSalt}"
```
hemant0308, line 115: i tried it, referencing the value from same file is not working. correct me if i am wrong. thats why added these values directly. @brig

server/dist/src/main/resources/concord-server.conf
```diff
@@ -112,7 +112,9 @@ concord-server {
...
+            # plain values of secretstore.secretStoreSalt and secretstore.serverPassword. using these values for migration of secrets.
+            #secretStoreSalt = "q1"
```
brig, line 116: why not use `${secretStore.secretStoreSalt}` and `${secretstore.serverPassword}` as values? 
secretStoreSalt/serverPassword is required and should already be defined in the cfg, so user doesn't have to do anything.

server/dist/src/main/resources/concord-server.conf
brig: hm... interesting. is `secretStore.secretStoreSalt` is defined in your cfg? share error?

brig: I think the problem is `""` :(

hemant0308: updated.

brig (approved):

PR #694, opened 2022-12-22
server-db: secret salt default value as decoded base64 value

brig:
when the value in quotes, it is used as is.


hemant0308:
LGTM. Tested locally

ibodrov (approved):

PR #699, opened 2023-01-03
license: copyright-up

amithkb:


brig (approved):

amithkb:
closing with a note that it is developer's responsibility to update the copyright year(if outdated) in the respective source files only whenever there is a content change.

PR #701, opened 2023-01-11
server: added 'last updated at' field for kv records

brig:


server/db/src/main/resources/com/walmartlabs/concord/server/db/liquibase.xml
```diff
@@ -106,5 +106,6 @@
     <include file="v1.95.0.xml" relativeToChangelogFile="true"/>
     <include file="v1.96.0.xml" relativeToChangelogFile="true"/>
     <include file="v1.98.1.xml" relativeToChangelogFile="true"/>
+    <include file="v1.99.xml" relativeToChangelogFile="true"/>
```
benbroadaway, line 109: Should be `v1.99.0.xml`, right?

brig: oops. thanks.

ibodrov (approved):

PR #705, opened 2023-01-17
concord-agent: make more delays configurable

ibodrov:
Configurable delays for successful and unsuccessful (connection errors) polling attempts.


ibodrov:
Don't merge it yet, looks like most of the re-connection logic is in the queue client and that is where I should add a config parameter.

ibodrov:
There we go.

brig (approved):

PR #707, opened 2023-01-23
server: use clean directory for each refresh listener

benbroadaway:
When a `RepositoryRefreshListener` processes the repo directory, it typically will process `imports` which adds files to the directory. If any of the files are symlinks, then multiple listeners re-processing the same directory will try to recreate existing symlinks causing an exception.

This change gives each listener a fresh copy of the base exported repo directory so the listener can do whatever it wants without assumptions.

Alternatively, we could refactor one of two things:
1.  Change [`IOUtils.copy(...)`](https://github.com/walmartlabs/concord/blob/53db9edaa44b0a028af57046dc7f57cfd87d801c/common/src/main/java/com/walmartlabs/concord/common/IOUtils.java#L253-L268) to delete and re-create any existing symlinks it encounters. That _may_ have some side effects.
2. Refactor the repository export process to not re-process imports after they've already been handled once.

### DIFF
--- a/it/server/src/test/java/com/walmartlabs/concord/it/server/ExternalImportsIT.java
+++ b/it/server/src/test/java/com/walmartlabs/concord/it/server/ExternalImportsIT.java
@@ -33,6 +33,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
+import java.util.function.Consumer;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


brig:
maybe just load Process Definition once and pass it to listeners, instead of load process definition from each listener?

listener-1:
https://github.com/walmartlabs/concord/blob/master/server/impl/src/main/java/com/walmartlabs/concord/server/repository/listeners/TriggerRefreshListener.java#L67

listener-2:
https://github.com/walmartlabs/concord/blob/master/server/impl/src/main/java/com/walmartlabs/concord/server/repository/listeners/ProcessDefinitionRefreshListener.java#L60

benbroadaway:
> maybe just load Process Definition once and pass it to listeners, instead of load process definition from each listener?
> 
> listener-1: https://github.com/walmartlabs/concord/blob/master/server/impl/src/main/java/com/walmartlabs/concord/server/repository/listeners/TriggerRefreshListener.java#L67
> 
> listener-2: https://github.com/walmartlabs/concord/blob/master/server/impl/src/main/java/com/walmartlabs/concord/server/repository/listeners/ProcessDefinitionRefreshListener.java#L60

I like that idea. Pushed a change which handles that. One question...should we pass the repo directory at all? The currently listeners don't need it now (there's one log statement which I didn't change which uses it). It's possible some future change may _want_ it.

brig (approved):

PR #709, opened 2023-01-28
feat(runtime): up level graal vm js engine lang level

cdaringe:
# problem

The current concord js primitives are painfully limited. Common, expected js idioms crash due to lack of support. JS users have expected some features to be present for nearly a decade in all common JS runtimes, but are not available in current provisions. For example, common methods, e.g. `Object.entries(...)` are not available. `.entries` is an interesting example, because it's actually associated with 2017 ECMAScript spec, but developers have add access to it since... 2014? due to wide availability _and_ pervasive downleveling of code in most JS compiler toolchains.

# solution

graal supports common, sensible, newer javascript in a fully backwards compatible way.

demo of newer syntax & methods, from a concord run with the new setting:
image

# refs

https://www.graalvm.org/latest/reference-manual/js/Options/


ibodrov (approved):
No objections from me.
@walmartlabs to decide if it should be a configuration parameter or not.

benbroadaway:
> No objections from me.
> 
> @walmartlabs to decide if it should be a configuration parameter or not.

We'll need to check for likelihood of breaking existing scripts. Probably needs to be something configurable so we don't have to roll back an upgrade just to flip it back.

brig:
+1 for configuration.

but maybe put configuration to script step? something like:
```
- script: js
  options:
      - js.ecmascript-version: "2020"
  body:
     ....
```
have no idea what to do with this options if we will try to change graalvm to something else, but why not:)

ibodrov:
-1 for exposing the JS engine options to users. If we ever migrate to another engine again, it might cause even more breakage. IMO, it should be a policy or a config in the agent. I think, the only purpose of this config is to provide a way to canary test the changes.

brig:
> -1 for exposing the JS engine options to users. If we ever migrate to another engine again, it might cause even more breakage. IMO, it should be a policy or a config in the agent. I think the only purpose of this config is to provide a way to canary test the changes.

yep-yep

cdaringe:
Lang level i think ought merit special consideration for exposure.

consider this: JS language levels are fully backwards compatible. They must be, or websites around the world start crashing as engines silently upgrade in browsers, which is explicitly considered by ECMAScript and TC39. The browser engine lang level upgrade scenario is the same scenario discussed here for concord. Runtime modifications, for which nashorn applies _many_, will surely will yield all of the pain during migration. If just the lang level is exposed, setting a max level could hedge most risk. That is, if a maximum value of ES2018 or ES2020 is applied, then the delta between the max allowed spec year and the current year should give years of buffer from hypothetical competing engines to “catch up”. If such an engine doesn't support a high enough language level, it’s probably not worth migrating to. If a mitigation is required and there’s no engine to turn to, drop js support in a major version bump. I love using JS, but one could argue “whats the point of offering JS if JS users cant use JS.”

Just my 2cents. Coach me! Ill do as the team requests, one way or another.

cdaringe:
Hey all,

- i've added another commit that adds `ecmascript-level` as an option
   - it's _opt in only_, and it's **just this singular engine option.**
- re: "we shouldn't expose engine options"
   - we could delete it in v3, if desired, and during said major release, update graal to a sane lang level default, rather than the current 1999 version of JS
      - again, i'd wager 99+% of JS expressions still work. re: "ecmascript is extremely backwards compatible"

thus, my proposal is, allow an opt-in flag now, purge it (or not) in the next concord release?

In this regard, people can use it now & it mitigates the risk of impacting other users. I'd be happy to re-label it as "experimental" or something, but would rely on your coaching :)

cdaringe:
kind and cordial bump. it's been a couple of weeks. thanks!

ibodrov:
@walmartlabs/concord any objections to merge?

benbroadaway:
> @walmartlabs/concord any objections to merge?

I was going to complain that this pr is only targeting runtime-v2....but that may actually be a good thing. the crustiest of old scripts are all going to be stuck in runtime-v1.

```yaml
- script: js
  in:
    ecmascript-version: "2022"
  body: |
    ...
```
I have some reservation about the implementation as a script `in:` variable. It's pretty loosey goosey and dashes in keys aren't great. But if we're considering this a temporary thing for opt-in testing then it's not the end of the world.

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
```diff
@@ -114,21 +114,34 @@ public String getLanguage(String languageOrExtension) {
...
 
+    private org.graalvm.polyglot.Context.Builder getGraalEngineContextBuilder(Map<String,Object> variables) {
+      HostAccess access = HostAccess.newBuilder(HostAccess.ALL)
+        .targetTypeMapping(Value.class, Object.class, Value::hasArrayElements, v -> new LinkedList<>(v.as(List.class))).build();
+      org.graalvm.polyglot.Context.Builder ctx = org.graalvm.polyglot.Context.newBuilder("js")
+        .allowHostAccess(access);
+      for (String key : variables.keySet()) {
+        switch (key) {
+          case "ecmascript-version":
+            ctx.option("js."+key, variables.get(key).toString());
```
benbroadaway, line 125: should we allow arbitrary values to get passed straight through here? Seems like it should come from a pre-existing list of options or at least go through some validation. It _should_ bonk when graalvm doesn't accept the value and gives a better-than-nothing error
```text
Error: For input string: "2022z"
```

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
```diff
@@ -114,21 +114,34 @@ public String getLanguage(String languageOrExtension) {
...
+            break;
+          default:
+            throw new RuntimeException("unsupported engine option: " + key);
```
benbroadaway, line 128: This should just ignore the invalid options rather than hijacking the entirety of input variables. E.g. this is now broken but should work
```yaml
- script: js
  in:
    myString: 'hello world'
  body: |
    print(myString);
````
results in
```text
Error: unsupported engine option: myString
```

benbroadaway (changes requested):

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
benbroadaway: ...apparently we need a test to make sure this isn't accidentally broken

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
```diff
@@ -114,21 +114,34 @@ public String getLanguage(String languageOrExtension) {
         return null;
     }
 
+    private org.graalvm.polyglot.Context.Builder getGraalEngineContextBuilder(Map<String,Object> variables) {
+      HostAccess access = HostAccess.newBuilder(HostAccess.ALL)
```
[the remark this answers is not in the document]
cdaringe: cool, will give that a shot.

do you know if there's something like https://github.com/jhipster/prettier-java, but more well known in the java community? most projects i work in now all have precommit hooks to eliminate formatting outright in all reviews, as there is only a single "correct" format. could be a nice add to the project!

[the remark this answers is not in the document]
cdaringe: 19c7f815

image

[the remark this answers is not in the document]
cdaringe: a84d8d55

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
```diff
@@ -114,21 +114,34 @@ public String getLanguage(String languageOrExtension) {
...
+            ctx.option("js."+key, variables.get(key).toString());
+            break;
+          default:
+            throw new RuntimeException("unsupported engine option: " + key);
+        }
+      }
+      return ctx;
+    }
+
     @SuppressWarnings({"rawtypes", "unchecked"})
```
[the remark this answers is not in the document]
cdaringe: d373d95d7

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
```diff
@@ -114,21 +114,34 @@ public String getLanguage(String languageOrExtension) {
         return null;
     }
 
+    private org.graalvm.polyglot.Context.Builder getGraalEngineContextBuilder(Map<String,Object> variables) {
+      HostAccess access = HostAccess.newBuilder(HostAccess.ALL)
+        .targetTypeMapping(Value.class, Object.class, Value::hasArrayElements, v -> new LinkedList<>(v.as(List.class))).build();
+      org.graalvm.polyglot.Context.Builder ctx = org.graalvm.polyglot.Context.newBuilder("js")
+        .allowHostAccess(access);
+      for (String key : variables.keySet()) {
```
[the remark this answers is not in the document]
cdaringe: d373d95d7

ibodrov (commented):
There's lots of formatting changes. Concord uses the default IDEA formatting rules, please re-format.

cdaringe:
> There's lots of formatting changes. Concord uses the default IDEA formatting rules, please re-format.

I've used both:

1. the hotkey that ben shared, and
2. the context menu

both produce the same results

image

i'm not a java nor intellij common user, so forgive me, but afaik i'm doing the obvious thing. what could i be missing?

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
cdaringe: yep, absolutely we should do that.

see: 9b6bc101e

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
cdaringe: removal of the throw: 97df51573 

test: yes, i'd love to write a test or two! if it's not too much of a burden, can one of you all send me a _seed_ or ideas here? given that this is firing up a _VM_, i suspect there will be some strong opinions on how such a test is setup. is it a unit test with graal mocked, is it an integration test where we actually execute graal?

i'm happy to work on the cases, but seems like some teamwork may be warranted for the initial setup?

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
```diff
@@ -37,153 +37,184 @@
 import java.io.BufferedWriter;
 import java.io.Reader;
 import java.io.Writer;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
```
cdaringe, line 43: intellij did this, not i!

cdaringe:
> dashes in keys aren't great

replaced with `esLevel` instead. i don't recall if this team is biased towards `camelCase` or `snake_case` options.

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
benbroadaway: depending on what needs to be tested, we have unit tests and integration tests that can handle most situations. Usual testing rules of thumb apply...all else equal, unit tests are preferable to integration tests. The ITs leverage [testcontainers](https://github.com/concord-workflow/testcontainers-concord) to test against a "real" concord instance.

> given that this is firing up a VM
For this specific instance, there isn't any extra VM being created. The default runtime covers the necessary dependencies.

There are existing tests similar to what needs to be done here. For ref:
 https://github.com/walmartlabs/concord/blob/68d57f52f412c94546815a5deaf8629bf9e9a3c5/runtime/v2/runner/src/test/java/com/walmartlabs/concord/runtime/v2/runner/MainTest.java#L437-L447

https://github.com/walmartlabs/concord/blob/68d57f52f412c94546815a5deaf8629bf9e9a3c5/runtime/v2/runner/src/test/resources/com/walmartlabs/concord/runtime/v2/runner/scriptVariablesSanitize/concord.yml#L1-L8

Add a similar test resource `<dir>/concord.yml` and a test in `MainTest`.

cdaringe: ah, thanks! very helpful. i don't know how i missed those! a few tests sprinkled in:

3beb9d4ef

cdaringe: realized I kept making JS errors and having formatting problems inline. moved that JS to a script file: 5b639ea13

cdaringe:
`MainTests.java` also formatted pretty drastically by intellij autoformatting.

review is very easy by using GitHub's `?w=1` flag:

https://github.com/walmartlabs/concord/pull/709/files?w=1

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
```diff
@@ -37,153 +37,184 @@
 import java.io.BufferedWriter;
```
cdaringe, line 37: TOP OF REVIEW - stop! review using GitHub whitespace ignore mode:

https://github.com/walmartlabs/concord/pull/709/files?w=1

brig (dismissed):
LGTM.

you can simplify the processing of in parameters a little, but we can do this separately :)

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/script/DefaultScriptEvaluator.java
ibodrov: Check if you're using the default code style:
image

Some plugins might also interfere here.

ibodrov:
This cannot be merged until the formatting is fixed. Please use the default code style in IDEA.

cdaringe:
> This cannot be merged until the formatting is fixed. Please use the default code style in IDEA.

editorconfig plugin was interfering

brig (approved):

benbroadaway (approved):

cdaringe:
how can i access the logs for the failed `its` jobs?

PR #711, opened 2023-01-30
feat: init easy script Variables instantiation

cdaringe:
# Problem

- `Variables` is cumbersome to work with from a scripting context

example:

```js
const task = tasks.get("slack");
const slackInputs = { foo: ...., bar: ...};
for (var key in slackInputs) {
  context.variables().set(key, slackInputs[key]);
}
task.execute(context.variables());
```

# Solution

Make working with variables more ergonomic/~~pure~~/declaritive/whatever by means of a fluent/chained `.withMap(...)` method

```js
```js
const task = tasks.get("slack");
task.execute(context.variables().withMap({ foo: ...., bar: ...}));
```

# Screenshot

(errr GH is having cert issues ATM with its static asset service :( )

### DIFF
--- a/runtime/v2/sdk/src/main/java/com/walmartlabs/concord/runtime/v2/sdk/Variables.java
+++ b/runtime/v2/sdk/src/main/java/com/walmartlabs/concord/runtime/v2/sdk/Variables.java
@@ -9,9 +9,9 @@ package com.walmartlabs.concord.runtime.v2.sdk;
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


benbroadaway:
> Make working with variables more ergonomic/pure/declaritive/whatever by means of a fluent/chained `.withMap(...)` method

I don't think this qualifies a "pure" function because setting variables from the execution context like this has the side effect of returning those variables back to the main process context. Which will be a bit sloppy, returning unexpected data back which may include sensitive data that shouldn't have snuck out. The script _should_ be creating a whole new `Variables` object to prevent this
```js
var MapBackedVariables = com.walmartlabs.concord.runtime.v2.sdk.MapBackedVariables;
const task = tasks.get("slack");
slack.execute(new MapBackedVariables({ foo: ..., bar: ...}));
```

cdaringe:
Totally, that was an erroneous claim. What you proposed is what i originally did… but IIRC all plugins write to the passed Variables, so i undid that impl, thus undid the pure soln

cdaringe:
this was my immutable attempt from a few days ago. i'm not our best java guy--https://github.com/walmartlabs/concord/compare/master...cdaringe:concord:feat/easier-variables-script-instantiation-2?expand=1

the objective was to say "ok, given whatever the backing impl of Variables is, make a copy of it..." but that may have been naive as well. i need to understand how concord actually resolves variables from context

PR #716, opened 2023-02-09
User disabled date

brig:


server/dist/src/main/resources/concord-server.conf
```diff
@@ -450,6 +450,9 @@ concord-server {
 
         # interval for the automatic group sync (interval)
         minAgeSync = "1 day"
+
+        # interval to delete disabled users (interval)
```
amithkb, line 454: If we want to disable `delete` old user.. then `disabledAge` should be set to some large?

ibodrov: That's actually a good question. I also don't think that this feature should be enabled by default.

brig: > If we want to disable delete old user.. then disabledAge should be set to some large one?

just set `disabledAge` to null...

brig: > That's actually a good question. I also don't think that this feature should be enabled by default.
fixed. now this functionality disabled by default...

ibodrov (approved):

PR #717, opened 2023-02-14
Fix Docker image build on aarch64 hosts

ibodrov:


docker-images/base/oss/debian/Dockerfile
```diff
@@ -17,12 +17,13 @@ RUN apt-get update && \
 ADD --chmod=755 ./get_jdk_url.sh /tmp/
 ARG jdk_version
```
amithkb, line 18: missing `ADD --chmod=755 ./get_arch.sh /tmp/`

ibodrov: thanks! I haven't tested the debian images yet, still working on getting CentOS images to pass the tests.

benbroadaway (approved):

PR #753, opened 2023-05-12
server: allow groupBy event attributes in GH triggers

brig:
for `pull_request` `groupBy: branch` is base branch, and it will be the same for all pull_requests to main/master...

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/events/GithubEventResource.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/events/GithubEventResource.java
@@ -23,6 +23,7 @@ package com.walmartlabs.concord.server.events;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.walmartlabs.concord.common.ConfigurationUtils;
 import com.walmartlabs.concord.runtime.v2.model.GithubTriggerExclusiveMode;
 import com.walmartlabs.concord.sdk.Constants;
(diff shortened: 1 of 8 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/events/GithubEventResource.java
```diff
@@ -193,18 +194,30 @@ public Map<String, Object> resolve(TriggerEntry t) {
...
+                if (payloadPath.length == 1) {
+                    throw new IllegalArgumentException("Invalid groupBy: '" + e.groupBy() + "'");
                 }
-                default:
-                    throw new IllegalArgumentException("Unknown groupBy: '" + e.groupBy() + "'");
+
+                payloadPath = Arrays.copyOfRange(payloadPath, 1, payloadPath.length);
+                Object maybeString = ConfigurationUtils.get(payload.raw(), payloadPath);
+                if (maybeString == null || (maybeString instanceof String)) {
+                    group = (String) maybeString;
```
benbroadaway, line 214: I'm trying to wrap my head around what happens when the value is `null`. Does that effectively not group it? Or is `null` a valid group and possibly a bunch of processes get grouped (and `cancel`led or `wait`ed)?

brig: logic will be the same as before, if group is null then no exclusive attributes for process:

https://github.com/walmartlabs/concord/blob/brig/gh-trigger-groupby-event-attr/server/impl/src/main/java/com/walmartlabs/concord/server/events/GithubEventResource.java#L223

ibodrov (approved):

PR #754, opened 2023-05-16
runtime-v2: mock tasks

brig:
Experimental feature, but might be useful :)


ibodrov:
Let's decide if we also need `assert`-like facilities here as well. Mocking tasks is usually a one part of a bigger problem -- flow unit testing. Perhaps we need to provide a more complete "concord-testing" plugin?

ibodrov (approved):
Asserts can be added later if needed. This LGTM as it is now.

brig:
docs: https://github.com/walmartlabs/concord-website/pull/168

PR #765, opened 2023-06-01
runtime-v2: error location for loop, call, parallel, retry commands

brig:
for expressions in step `name`, `in/out/items` variables.

for example. before:
```
19:47:20 [INFO ] Process status: RUNNING
19:47:20 [ERROR] Can't find a variable 'undefined' used in '${undefined}'. Check if it is defined in the current scope. Details: ELResolver cannot handle a null base Object with identifier 'undefined'
19:47:20 [ERROR] Process exit code: 1
19:47:20 [INFO ] Process status: FAILED
```

after:
```
20:19:26 [INFO ] Process status: RUNNING
20:19:26 [ERROR] (concord.yml): Error @ line: 7, col: 7. Can't find a variable 'undefined' used in '${undefined}'. Check if it is defined in the current scope. Details: ELResolver cannot handle a null base Object with identifier 'undefined'
20:19:27 [ERROR] Process exit code: 1
20:19:27 [INFO ] Process status: FAILED
```


benbroadaway (approved):

ibodrov:
The error handling in v2 in general needs another design pass. We need to provide ways to handle errors for individual commands, while evaluating arguments, and for "wrapper" commands like loops and retries.

PR #772, opened 2023-06-27
concord-ansible-plugin: fix handling of play and task names longer than 1024 chars

benbroadaway:
We have a limit of 1024 chars for the value of play and task names for ansible playbook stats. The event processor currently attempts to insert any length of text given from the playbook execution and fails to progress. This fix truncates the name to the max length.


ibodrov (approved):

server/plugins/ansible/impl/src/main/java/com/walmartlabs/concord/server/plugins/ansible/PlayInfoProcessor.java
```diff
@@ -83,6 +83,10 @@ public void process(DSLContext tx, List<Event> events) {
                     continue;
                 }
 
+                if (playName.length() > 1024) {
```
brig, line 86: we can use something like:
```
ANSIBLE_PLAY_STATS.PLAY_NAME.getDataType().length()
```
instead of 1024 constant

benbroadaway: That's a much better idea.

brig (approved):

PR #774, opened 2023-06-28
runtime-v2: hasNonNullVariable function

brig:
instead of:
`${hasVariable('var') && var != null}

we can use:
`${hasNonNullVariable('var')}`


benbroadaway (approved):

brig:
docs: https://github.com/walmartlabs/concord-website/pull/167

PR #775, opened 2023-06-29
console2: secrets lastUpdatedAt and createdAt

brig:
Screenshot 2023-07-03 at 14 36 13


console2/src/components/molecules/HumanizedDuration/index.tsx
```diff
@@ -39,10 +51,9 @@ export default class extends React.PureComponent<Props> {
         return (
             <Popup trigger={<span>{s}</span>}>
```
benbroadaway, line 52: What do you think about putting a two or five (or more?) second delay on this popup? Or some other mechanism that allows interaction with the popup. Currently, it's not possible to copy the value in the popup because it disappears as soon as the mouse leaves the element

brig: added `hoverable={true}` to some popups :)

benbroadaway (approved):

PR #791, opened 2023-07-31
runtime-v2: new varibales scope

brig:


ibodrov:
This is a change that can break a lot of existing flows. We might consider this for v3 or try solving it with linters and a combination of static and runtime analysis.

PR #808, opened 2023-09-10
new activity page

brig:
Screenshot 2023-09-09 at 22 53 37 (1)


console2/src/api/service/console/user/index.ts
```diff
@@ -18,30 +18,34 @@
...
 ): Promise<UserActivity> =>
     fetchJson(
-        `/api/service/console/user/activity?${queryParams({ maxProjectsPerOrg, maxOwnProcesses })}`
+        `/api/v2/service/console/user/activity?${queryParams({ maxOwnProcesses })}`
     );
+
+export const listProcessCards = (
+): Promise<ProcessCardEntry[]> =>
+    fetchJson(
+        `/api/v2/service/console/user/process-card`
```
ibodrov, line 50: Should it be `/api/v1/org/process-card` instead?
On the same note, I think we should use `/api/v1/processCard` prefix instead of putting it under `/api/v1/org` (the next URL part is taken by `{orgId}` anyway).

ibodrov (approved):

PR #810, opened 2023-09-20
concord client v2

brig:
new concord client impl (client2):
- standard java http client without additional dependencies (jackson only:));
- removed all custom HTTP requests, so we can now use generated API classes;

This PR is just to release Concord with the new client. After that, we'll build concord-testcontainers with the new Concord version and in additional PR we will fully switch to client2 in Concord.


client2/pom.xml
```diff
@@ -0,0 +1,174 @@
...
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.revapi</groupId>
+                <artifactId>revapi-maven-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                    <analysisConfiguration>
+                        <revapi.ignore>
```
ibodrov, line 158: Do we need ignores for a new module?

client2/pom.xml
```diff
@@ -0,0 +1,174 @@
...
+
+        <!-- Immutables -->
+        <dependency>
+            <groupId>org.immutables</groupId>
+            <artifactId>value</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
```
ibodrov, line 58: This and `error_prone_annotations` below -- not used in the current client code, if I read it correctly?

ibodrov (approved):
Minor comments, otherwise LGTM.

PR #844, opened 2023-12-08
runtime-v2: allow "true|false" string in if expression

brig:
useful when flow called with cli...

### DIFF
--- a/runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/vm/IfCommand.java
+++ b/runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/vm/IfCommand.java
@@ -54,11 +54,29 @@ public class IfCommand extends StepCommand<IfStep> {
         EvalContext evalContext = ecf.global(ctx);
 
         ExpressionEvaluator ee = runtime.getService(ExpressionEvaluator.class);
-        Boolean ifResult = ee.eval(evalContext, expr, Boolean.class);
-        if (ifResult != null && ifResult) {
+        Object ifResult = ee.eval(evalContext, expr, Object.class);
+        if (isTrue(ifResult)) {
             frame.push(thenCommand);
         } else if (elseCommand != null) {
             frame.push(elseCommand);
         }
+    }
+
+    private static boolean isTrue(Object value) {
+        if (value == null) {
+            return false;
+        }
+
+        if (value instanceof Boolean b) {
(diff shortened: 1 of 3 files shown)


runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/vm/IfCommand.java
```diff
@@ -54,11 +54,29 @@ protected void execute(Runtime runtime, State state, ThreadId threadId) {
...
+
+    private static boolean isTrue(Object value) {
+        if (value == null) {
+            return false;
+        }
+
+        if (value instanceof Boolean b) {
+            return b;
+        } else if (value instanceof String s) {
+            if ("true".equalsIgnoreCase(s)) {
```
ibodrov, line 73: Do we really need to coerce string values "true/false" to boolean? I don't think we do that anywhere else?

ibodrov:
Perhaps the CLI should parse the input values as JSON instead. E.g. treat `true` or `false` as boolean values and `"true"` and `"false"` as strings?

brig:
sure. better to implement it in cli. I don't really like this PR either :)

but what I don't like even more is that
```if: ${"true" == true}``` is true and ok, and ````${"true"}``` is the class cast exception :(

**Comment by ibodrov** [comment]
Yeah, I can't really think of a better way for now. Let's keep it that way? I took a stab at the docs https://github.com/walmartlabs/concord-website/pull/158/files

**Comment by brig** [comment]
more working examples:)
```
    - if: ${myString == true}
      then:
        - log: "Yep!"
      else:
        - throw: "Unexpected"

    - if: ${true == true}
      then:
        - log: "Yep!"
      else:
        - throw: "Unexpected"

    - if: ${!myString == false}
      then:
        - log: "Yep!"
      else:
        - throw: "Unexpected"

    - if: ${!"true" == false}
      then:
        - log: "Yep!"
      else:
        - throw: "Unexpected"

    - expr: "${myString ? true : false}"
      out: r1

    - if: "${r1}"
      then:
        - log: "Yep!"
      else:
        - throw: "Unexpected"
```

logs:
```
Starting...
15:09:48.301 [main] Yep!
15:09:48.303 [main] Yep!
15:09:48.303 [main] Yep!
15:09:48.304 [main] Yep!
15:09:48.305 [main] Yep!
...done!
```

ibodrov (approved):

PR #850, opened 2023-12-18
server, console2: ability to restart the runtime-v2 process

brig:
Screenshot 2023-12-18 at 13 37 39

the process can only be restarted from the following final statuses:
```
CANCELLED 
FAILED
FINISHED
TIMED_OUT
```

Only initiator of the process or admin can restart it.

User can only restart the root process. if a **child** process is restarted, then the **root** process is restarted.

A new table `PROCESS_INITIAL_STATE` has been added for storing the initial state of the process. This way, it's possible to separately regulate the space occupied in the database (with partitions cleanup script). Plus, there's one less insert/delete in the `PROCESS_STATE_TABLE` table :)

next step will be to display logs only for a specific attempt.

restart process popup:
Screenshot 2024-01-12 at 01 38 43


console2/src/components/organisms/RestartProcessPopup/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    const { trigger, refresh } = props;
+
+    return (
+        <SingleOperationPopup
+            trigger={trigger}
+            title="Restart the process?"
+            introMsg={
+            <>
+                <p>Are you sure you want to restart the selected process?</p>
```
ibodrov, line 72: Something like this, maybe.

```suggestion
{!parentInstanceId && <p>Are you sure you want to restart the selected process?</p>}
{parentInstanceId &&
<Message warning>
You are about to restart a child process. Please note:

    Only the <a href="...">parent (root) process</a> will be restarted.
    Restarting the parent process may also re-run this and other child processes.
    Ensure that this is the desired action before proceeding.

Do you want to continue with restarting the parent process?
</Message>}
```

ibodrov (approved):

PR #871, opened 2024-02-05
runtime-v2: ignore empty string as sensitive data

brig:


benbroadaway (approved):

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/SensitiveDataHolder.java
```diff
@@ -43,6 +43,10 @@ public boolean empty() {
     }
 
     public void add(String sensitiveData) {
+        if (sensitiveData == null || sensitiveData.trim().isEmpty()) {
```
ibodrov, line 46: There's also String#isBlank (JDK 11+, I believe)

ibodrov (approved):

ibodrov (approved):

PR #872, opened 2024-02-09
server: skip repository refresh when repo is disabled

benbroadaway:
Currently, a repo cannot be disabled if there's something wrong with the config (auth, branch/tag, path etc).


dankle (approved):

dankle (approved):

mtkunkel (approved):

benbroadaway:
need to double-check cron triggers are cleared out appropriately...

benbroadaway:
> need to double-check cron triggers are cleared out appropriately...

👍👍 `trigger_schedule` is cleaned up with cascading deletes from the [`triggers` table](https://github.com/walmartlabs/concord/blob/0656ba5b3a3d1ffb44fb2851d8d106a5f1664e61/server/db/src/main/resources/com/walmartlabs/concord/server/db/v0.65.0.xml#L24)

ibodrov (approved):

PR #880, opened 2024-03-06
agent: kill runner child PIDs

benbroadaway:
Funky child processes from the runner may persist even after the runner is killed due to cancel command or `processTimeout`. This change kills each child recursively if they are still alive after the runner is killed.

### DIFF
--- a/agent/src/main/java/com/walmartlabs/concord/agent/Utils.java
+++ b/agent/src/main/java/com/walmartlabs/concord/agent/Utils.java
@@ -23,30 +23,48 @@ package com.walmartlabs.concord.agent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.lang.reflect.Field;
+import java.util.List;
 
 public final class Utils {
 
     private static final Logger log = LoggerFactory.getLogger(Utils.class);
 
     public static boolean kill(Process proc) {
-        if (!proc.isAlive()) {
+        return kill(proc.toHandle());
+    }
+
+    public static boolean kill(Process proc, boolean killDescendants) {
+        List<ProcessHandle> children = killDescendants ? List.of() : proc.children().toList();
+
+        // kill parent first which may gracefully clean up all descendents
+        boolean killed = kill(proc.toHandle());
+
+        // clean up orphaned processes that are still running
+        children.stream()
+                .flatMap(ProcessHandle::descendants)
+                .forEach(Utils::kill);
+
+        return killed;
+    }
+
+    private static boolean kill(ProcessHandle handle) {
+        if (!handle.isAlive()) {
             return false;
         }
 
-        String p = toString(proc);
+        String p = toString(handle);
 
         log.info("kill ['{}'] -> attempting to stop...", p);
-        proc.destroy();
+        handle.destroy();
 
-        if (proc.isAlive()) {
+        if (handle.isAlive()) {
             sleep(1000);
         }
 
-        while (proc.isAlive()) {
+        while (handle.isAlive()) {
             log.warn("kill ['{}'] -> waiting for the process to die...", p);
             sleep(3000);
-            proc.destroyForcibly();
+            handle.destroyForcibly();
         }
 
         return true;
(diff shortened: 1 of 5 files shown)


dankle (dismissed):

agent/src/main/java/com/walmartlabs/concord/agent/Utils.java
```diff
@@ -23,30 +23,48 @@
...
+    public static boolean kill(Process proc, boolean killDescendants) {
+        List<ProcessHandle> children = killDescendants ? List.of() : proc.children().toList();
+
+        // kill parent first which may gracefully clean up all descendents
+        boolean killed = kill(proc.toHandle());
+
+        // clean up orphaned processes that are still running
+        children.stream()
+                .flatMap(ProcessHandle::descendants)
+                .forEach(Utils::kill);
```
ibodrov, line 45: If we got a lot of those "funky" processes, which might not respond to `handle.destroy()` immediately, does that means an occasional >=1s `sleep` calls for each child?
Should use shorter sleeps in `kill` then?

ibodrov (approved):
LGTM, minor Q

agent/src/main/java/com/walmartlabs/concord/agent/Utils.java
benbroadaway: gooood question. What i've observed is that typically it's one sub-process in a tree that's the holdup, but it certainly could be multiple in the same sub-process tree or elsewhere.

At least in a simple recreation of hte issue with an ansible playbook like
```yaml
- hosts: all
  tasks:
  - name: "Holdup"
    ansible.builtin.shell:
      executable: /bin/bash
      cmd: |
        trap '' SIGPIPE
        trap '' SIGTERM
        while true
        do
          sleep 1 # could do something here, or just sleep
        done
```

When the process is interrupted (timeout, cancel), currently it leaves behind these zombies
```text
root@8bdf8700ad0a:/workspace# ps -efj
UID        PID  PPID  PGID   SID  C STIME TTY          TIME CMD
concord   4130     9     9     8 73 22:41 ?        00:00:08 /usr/lib/jvm/zre-17-amd64/b...
concord   4160  4159     9     8 34 22:41 ?        00:00:02 /usr/bin/python3 /usr/local...
concord   4192  4160     9     8  5 22:41 ?        00:00:00 /usr/bin/python3 /usr/local...
concord   4202  4192     9     8  0 22:41 ?        00:00:00 /bin/sh -c /usr/bin/python3...
concord   4203  4202     9     8  3 22:41 ?        00:00:00 /usr/bin/python3 /tmp/conco...
concord   4204  4203     9     8  0 22:41 ?        00:00:00 /bin/bash -c trap '' SIGPIP...
```

and the agent has to do a >1s wait for the intentionally feisty one

```text
22:41:50.711 [pool-2-thread-9] [INFO ] com.walmartlabs.concord.agent.Utils - kill ['pid=4130'] -> attempting to stop...
22:41:51.712 [pool-2-thread-9] [INFO ] com.walmartlabs.concord.agent.Utils - kill ['pid=4160'] -> attempting to stop...
22:41:52.713 [pool-2-thread-9] [INFO ] com.walmartlabs.concord.agent.Utils - kill ['pid=4192'] -> attempting to stop...
22:41:53.716 [pool-2-thread-9] [INFO ] com.walmartlabs.concord.agent.Utils - kill ['pid=4202'] -> attempting to stop...
22:41:54.717 [pool-2-thread-9] [INFO ] com.walmartlabs.concord.agent.Utils - kill ['pid=4203'] -> attempting to stop...
22:41:55.719 [pool-2-thread-9] [INFO ] com.walmartlabs.concord.agent.Utils - kill ['pid=4204'] -> attempting to stop...
22:41:56.720 [pool-2-thread-9] [WARN ] com.walmartlabs.concord.agent.Utils - kill ['pid=4204'] -> waiting for the process to die...
22:41:59.722 [pool-2-thread-9] [WARN ] com.walmartlabs.concord.agent.Utils - kill ['pid=4204'] -> waiting for the process to die...
```
So if that was a 'real' playbook executing with 200 forks and one process from each fork was stuck, then yeah that'd be a minimum of 600 seconds of waiting

ibodrov: Thanks! I guess if it becomes a problem we can always rewrite `kill` to send SIGTERM to a bunch of Handles first then wait.

ibodrov (approved):

PR #882, opened 2024-03-09
fix trigger id calculation for complex args: heterogeneous lists, list of maps

brig:


server/impl/src/test/java/com/walmartlabs/concord/server/org/triggers/TriggerInternalIdCalculatorTest.java
```diff
@@ -61,4 +62,45 @@ public void test1() {
...
+    }
+
+    @Test
+    public void testNotEquals() {
+        String name = "trigger";
+        List<String> activeProfiles = new ArrayList<>();
+        Map<String, Object> arguments = new HashMap<>();
+        Map<String, Object> conditions = new HashMap<>();
+
+        Map<String, Object> cfg1 = new HashMap<>();
```
ibodrov, line 92: `Map.of` here should be less noisy, I think

ibodrov (dismissed):

ibodrov (approved):

PR #884, opened 2024-03-20
server: synchronous processing of process completion wait conditions

brig:


server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/XYZProcessStatusListener.java
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ * =====
+ */
+
+import com.walmartlabs.concord.server.process.queue.ProcessStatusListener;
+import com.walmartlabs.concord.server.sdk.ProcessKey;
+import com.walmartlabs.concord.server.sdk.ProcessStatus;
+import org.jooq.DSLContext;
+
+public class XYZProcessStatusListener implements ProcessStatusListener {
```
ibodrov, line 28: XYZ? Is it a temporary name?
Alternatively, `WaitConditionUpdater` maybe? Or `WaitConditionProcessStatusListener`.

ibodrov (dismissed):
A Q about naming, otherwise LGTM.

server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/XYZProcessStatusListener.java
brig: yep. thanks:)

ibodrov (approved):

PR #894, opened 2024-04-25
concord-server, tasks: disable repos on deleted ref, only refresh repos matching event branch

benbroadaway:
Two changes:

## Server: Disable repos when delete event is recieved

When the server handles a GitHub `push` event with `deleted: true`, then disable any repos which have a branch configured to match the event's `ref` value (i.e. the branch).

This cuts down on the number of useless `ConcordSystem` process which refresh broken repos.

## Repository Refresh Task: Skip repository refresh for repos that don't match `push` event's `ref`

There shouldn't be a need to refresh repos configured for other branches, tags, or commit IDs when only one branch was updated. So when a team configures dozes (or hundreds) of Concord repos tied to the same GitHub repo, but different branches, this cuts a lot of noise.


dankle (approved):

server/impl/src/main/java/com/walmartlabs/concord/server/events/github/GithubTriggerV2Processor.java
```diff
@@ -81,6 +102,26 @@ private void enrichEventConditions(Payload payload, TriggerEntry trigger, Map<St
         }
     }
 
+    private void disableRepo(RepositoryEntry repo, Payload payload) {
+        log.info("disable repo ['{}', '{}'] -> ref deleted", repo.getId(), payload.getBranch());
+        repoDao.disable(repo.getId());
```
ibodrov, line 107: I think if we should wrap these two calls into a transaction.

ibodrov (commented):
Still looking through it, a minor comment so far.

server/impl/src/main/java/com/walmartlabs/concord/server/events/github/GithubTriggerV2Processor.java
benbroadaway: For combining in a transaction, is it better to duplicate the [trigger delete](https://github.com/walmartlabs/concord/blob/60b237267753d4dbfc7feedb21e6e6641a439b46/server/impl/src/main/java/com/walmartlabs/concord/server/org/triggers/TriggersDao.java#L79-L83) code in `RepositoryDao`? or inject `TriggersDao` into `RepositoryDao`?

server/impl/src/main/java/com/walmartlabs/concord/server/org/project/RepositoryDao.java
```diff
@@ -38,12 +39,15 @@
 public class RepositoryDao extends AbstractDao {
 
     private final ConcordObjectMapper objectMapper;
+    private final TriggersDao triggersDao;
 
     @Inject
     public RepositoryDao(@MainDB Configuration cfg,
-                         ConcordObjectMapper objectMapper) {
+                         ConcordObjectMapper objectMapper,
```
ibodrov, line 46: I think it might be cleaner to make a small internal Dao class in GithubTriggerV2Processor and "compose" RepositoryDao and TriggersDao there instead of making a dependency here.

ibodrov (changes requested):

ibodrov (approved):

PR #897, opened 2024-04-30
runtime-v2: sensitive data in maps

brig:
check any `T get(Object key)` method as @sensitiveData method


runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/el/resolvers/SensitiveDataProcessor.java
```diff
@@ -24,13 +24,19 @@
...
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 public class SensitiveDataProcessor {
 
+    public static void process(Object value, List<Method> methods) {
+        Method method = methods.stream().filter(SensitiveDataProcessor::isSensitiveData).findFirst().orElse(null);
```
ibodrov, line 32: Why don't we want to call process for all methods we find here?

ibodrov (approved):
Adjusted naming a bit, otherwise LGTM

PR #903, opened 2024-05-09
server: fix wait conditions init after process restart

brig:


server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessStatusListener.java
```diff
@@ -52,9 +53,16 @@ public void onStatusChange(DSLContext tx, ProcessKey processKey, ProcessStatus s
     }
 
     private static void init(DSLContext tx, ProcessKey processKey) {
-        tx.insertInto(PROCESS_WAIT_CONDITIONS)
-                .columns(PROCESS_WAIT_CONDITIONS.INSTANCE_ID, PROCESS_WAIT_CONDITIONS.INSTANCE_CREATED_AT)
-                .values(processKey.getInstanceId(), processKey.getCreatedAt())
+        tx.insertInto(PROCESS_WAIT_CONDITIONS, PROCESS_WAIT_CONDITIONS.INSTANCE_ID, PROCESS_WAIT_CONDITIONS.INSTANCE_CREATED_AT)
```
ibodrov, line 56: Can we just do nothing here when `PREPARING`? If we skip the second insert anyway...

ibodrov (commented):
The fix looks fine, but I've got a Q.

server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessStatusListener.java
brig: all root processes starts with NEW status, and forks do not enter this status; they go directly to PREPARING. I don't remember why it was done this way...

ibodrov: Ah cool. Then we maybe we should use two separate SQLs here, one with the existence check for PREPARING and the old one for NEW? Saves us an index read?

brig: user submits process: 
NEW -> PREPARING (just DB update, without history, or listeners notify)-> ENQUEUED -> ... -> 

restart process:
NEW -> PREPARING (just DB update, without history, or listeners notify)-> ENQUEUED -> ... -> 

child process:
NEW -> PREPARING (just DB update, without history, or listeners notify)-> ENQUEUED -> ... -> 

fork process:
PREPARING -> ENQUEUED -> ... ->

ibodrov (approved):

PR #910, opened 2024-05-16
server: process wait conditions in batch mode

brig:
the idea here is this: instead of processing each wait condition individually, we will process them in batches. As a result, instead of requesting the status for each process separately, we will request the statuses for a batch of processes at once

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessFinishHandler.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessFinishHandler.java
@@ -41,6 +41,7 @@ import java.util.stream.Collectors;
 
 import static com.walmartlabs.concord.server.jooq.tables.ProcessQueue.PROCESS_QUEUE;
 import static com.walmartlabs.concord.server.process.waits.ProcessCompletionCondition.CompleteCondition;
+import static org.jooq.impl.DSL.any;
 
 /**
  * Handles the processes that are waiting for other processes to finish.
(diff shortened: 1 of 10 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessFinishHandler.java
```diff
@@ -52,16 +75,127 @@ public WaitType getType() {
...
-        if (wait.resumeEvent() != null) {
-            return Result.resume(wait.resumeEvent());
-        } else {
-            return Result.action(tx -> processQueueManager.updateExpectedStatus(tx, processKey, ProcessStatus.WAITING, ProcessStatus.ENQUEUED));
+        public Map<UUID, ProcessStatus> findStatuses(Set<UUID> processes) {
+            Map<UUID, ProcessStatus> result = txResult(tx -> {
+                ProcessQueue q = PROCESS_QUEUE.as("q");
+                return tx.select(q.INSTANCE_ID, q.CURRENT_STATUS)
+                        .from(q)
+                        .where(q.INSTANCE_ID.in(processes))
```
ibodrov, line 192: Is JOOQ smart enough to turn this `INSTANCE_ID IN (...)` into an array operation? I'm worried that it's going to fail for large `processes` sets.

ibodrov (commented):
Still looking through it, one Q so far.

server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessFinishHandler.java
brig: that's why we split processes here:
https://github.com/walmartlabs/concord/pull/910/files/8a2282d40442a02df930f0192f4dd1f2b5e6e391#diff-33569c60af8758f4910c839fbc4b973bd155e2cc59016ca55f0d9cfe354948a3R90

ibodrov: So the upper limit is `waitProcessLimitForStatusQuery`? Should be fine, I think, but for the query parser sake I'd replaced with an array operation if it is possible here.

brig: yes, `waitProcessLimitForStatusQuery` is a limit

ibodrov (approved):

PR #913, opened 2024-05-20
dependency-manager: allow `LATEST` to pull latest version from remote repositories

amithkb:
Example:

`mvn://com.walmartlabs.concord:concord-sdk:LATEST`

### DIFF
--- a/dependency-manager/src/main/java/com/walmartlabs/concord/dependencymanager/DependencyManager.java
+++ b/dependency-manager/src/main/java/com/walmartlabs/concord/dependencymanager/DependencyManager.java
@@ -75,6 +75,7 @@ public class DependencyManager {
 
     private static final String FILES_CACHE_DIR = "files";
     public static final String MAVEN_SCHEME = "mvn";
+    private static final String LATEST = "LATEST";
 
     private final Path cacheDir;
     private final Path localCacheDir;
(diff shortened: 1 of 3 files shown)


brig:
As far as I remember, LATEST should resolve without any changes, no?

works for me from master:
```
flows:
  default:
  - log: "Hello, ${name}!"

configuration:
  arguments:
    name: "Concord"
  debug: true
  dependencies:
    - "mvn://org.apache.james:apache-james-clamav:LATEST"
```

```
...
/.../.m2/repository/org/apache/james/apache-james-clamav/3.8.1/apache-james-clamav-3.8.1.jar
...
```

amithkb:
> As far as I remember, LATEST should resolve without any changes, no?
> 
> works for me from master:
> 
> ```
> flows:
>   default:
>   - log: "Hello, ${name}!"
> 
> configuration:
>   arguments:
>     name: "Concord"
>   debug: true
>   dependencies:
>     - "mvn://org.apache.james:apache-james-clamav:LATEST"
> ```
> 
> ```
> ...
> /.../.m2/repository/org/apache/james/apache-james-clamav/3.8.1/apache-james-clamav-3.8.1.jar
> ...
> ```

No Interestingly. By default aether looks for local maven repository for `LATEST`.

Scenario where it won't work:
- when remote repositories configured with no explicit `updatePolicy`. By default it is set to `daily` (check on the first run of the day). - [Reference](https://maven.apache.org/settings.html#:~:text=or%20snapshots).-,updatePolicy,-%3A%20This%20element%20specifies)
- Suppose during the day after the first run if any new version is updated in the remote. Subsequent maven resolution won't look for remote for new versions.
- So solution is to set `updatePolicy` to `always` incase of `LATEST` so that everytime it can check whether any new version is available to download(Note: It only downloads if local mvn repo doesn't consists of new version)

brig (dismissed):

ibodrov:
The approach looks good to me, didn't finish reviewing the code yet. One potential optimization that I'd like to see (eventually, doesn't have to be in the same PR) is to re-use a single MavenRepositorySession between calls.

amithkb:
> The approach looks good to me, didn't finish reviewing the code yet. One potential optimization that I'd like to see (eventually, doesn't have to be in the same PR) is to re-use a single MavenRepositorySession between calls.

reduced to single MavenRepositorySession call

ibodrov (approved):

brig (approved):

PR #917, opened 2024-05-30
runtime-v2: allow double values in yaml

brig:


ibodrov (approved):

brig:
Maybe it would be better to throw an exception if a double value is encountered, instead of just replacing it with null?

ibodrov:
I'd love to see us hard error-ing here but it might be too harsh for our current users.
Maybe in the runtime v3 we'll do that. :)

PR #919, opened 2024-06-06
runtime-v2: command that will always be executed

brig:
in theory we can define something like this in yaml:
```
- try:
   - log: "boo"
   - throw: "error"
  finally:
   - log: "cleanup"
```
right now it's just a command interface :)

### DIFF
--- a/runtime/v2/vm/src/main/java/com/walmartlabs/concord/svm/InMemoryState.java
+++ b/runtime/v2/vm/src/main/java/com/walmartlabs/concord/svm/InMemoryState.java
@@ -83,8 +83,29 @@ public class InMemoryState implements Serializable, State {
     }
 
     @Override
-    public void popFrame(ThreadId threadId) {
+    public void popFrame(ThreadId threadId, UnstoppableCommandHandler handler) {
         log.trace("popFrame {}", threadId);
+
+        Frame currentFrame = peekFrame(threadId);
+
+        // finalizers
+        Collection<Exception> causes = new ArrayList<>();
+
+        currentFrame.getCommands()
+                .stream()
+                .filter(c -> c instanceof UnstoppableCommand)
+                .map(c -> (UnstoppableCommand) c)
+                .forEach(c -> {
+                    try {
+                        handler.handle(c);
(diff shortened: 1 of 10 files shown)


runtime/v2/vm/src/main/java/com/walmartlabs/concord/svm/UnstoppableCommand.java
```diff
@@ -0,0 +1,29 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =====
+ */
+
+public interface UnstoppableCommand extends Command {
```
ibodrov, line 23: Can we call it `FinallyBlock`, `FrameFinalizer` or just `Finalizer` ?. "Unstoppable" is too dramatic to my taste. :-)

runtime/v2/vm/src/main/java/com/walmartlabs/concord/svm/InMemoryState.java
```diff
@@ -83,9 +83,30 @@ public Frame peekFrame(ThreadId threadId) {
...
-    public void popFrame(ThreadId threadId) {
+    public void popFrame(ThreadId threadId, UnstoppableCommandHandler handler) {
         log.trace("popFrame {}", threadId);
 
+        Frame currentFrame = peekFrame(threadId);
+
+        // finalizers
+        Collection<Exception> causes = new ArrayList<>();
+
+        currentFrame.getCommands()
```
ibodrov, line 94: We could store all the `finally` commands as a block command in a separate field in `Frame`, I think.
And fill it when we `build()` the frame. It should not be too different from the exceptionHandler there.

brig: will close this PR

PR #946, opened 2024-07-27
agent-operator: use agent maintenance mode before terminating agent

brig:


agent-operator/src/main/java/com/walmartlabs/concord/agentoperator/agent/DefaultAgentClient.java
```diff
@@ -0,0 +1,82 @@
...
+                .uri(new URI(url))
+                .POST(HttpRequest.BodyPublishers.noBody())
+                .timeout(REQUEST_TIMEOUT)
+                .build();
+
+        client.send(request, HttpResponse.BodyHandlers.ofString());
+    }
+
+    @Override
+    public boolean isNoWorkers() throws Exception {
```
ibodrov, line 60: Maybe reverse the method's logic here? 
```suggestion
    public boolean hasBusyWorkers() throws Exception {
```
Seeming that the only use of this method is a check for the reverse condition.

agent-operator/src/main/java/com/walmartlabs/concord/agentoperator/agent/AgentClientFactory.java
```diff
@@ -0,0 +1,55 @@
...
+                    .version(HttpClient.Version.HTTP_1_1)
+                    .followRedirects(NEVER)
+                    .connectTimeout(CONNECT_TIMEOUT)
+                    .build();
+        } else {
+            httpClient = null;
+        }
+    }
+
+    public AgentClient create(String podIp) {
```
ibodrov, line 48: Maybe pass the entire Pod object here? :thinking: 
It's not like we need anything but the IP here (for now) but it feels safer to avoid naked Strings.

ibodrov (approved):
Minor comments, otherwise LGTM

ibodrov (approved):

PR #949, opened 2024-08-04
runtime-v2: option for event batching for runner events

benbroadaway:
On a typical day, the fleet of runners may report single process events at rates of over 500 events per second. This is both relatively chatty keeping the server API busy juggling many connections and slows down runner execution because each event is reported synchronously.

This PR add a runner option to enable batching of events. Default size of `1` effectively behaves the same as existing runner. Enable batching by increasing the value.

When batching, process events are queued up to the max batch size before reporting the batch synchronously (the next process step does not execute until the batch is flushed). If a long-running step occurs (e.g. `ansible`), then a background timer will force a flush within the configured duration (default of `15` seconds). A flush also occurs when the process terminates (suspends, or final termination).


runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/DefaultEventReportingService.java
```diff
@@ -0,0 +1,154 @@
...
+            getProcessEventsApi().event(instanceId.getValue(), req);
+        } catch (ApiException e) {
+            log.warn("error while sending an event to the server: {}", e.getMessage());
+        }
+    }
+
+    /**
+     * @return batch of up-to {@link #maxBatchSize} queued process events
+     */
+    private List<ProcessEventRequest> takeBatch() {
```
ibodrov, line 125: I think this might be an alternative.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/BlockingQueue.html#drainTo(java.util.Collection,int)

benbroadaway: I'm doing some thinking about that. The only downside is we have to be really strict about not over-filling the queue since that results in an exception with, say `ArrayBlockingQueue`--which may possibly happen with my first stab here when `parallel` blocks are running. It still reports up-to exactly the max batch size, but sometimes chatty threads can fill it up with more than that many events. Perhaps that's worth reigning in anyway.

benbroadaway: ...or just use a `LinkedBlockingQueue` without a bound. Though I think I sorted out the possible routes of overfilling now.

ibodrov (approved):

PR #950, opened 2024-08-05
runtime-v1: option for event batching for runner events

benbroadaway:
Similar to https://github.com/walmartlabs/concord/pull/949 but for runtime-v1.

On a typical day, the fleet of runners may report single process events at rates of over 500 events per second. This is both relatively chatty keeping the server API busy juggling many connections and slows down runner execution because each event is reported synchronously.

This PR add a runner option to enable batching of events. Default size of 1 effectively behaves the same as existing runner. Enable batching by increasing the value.

When batching, process events are queued up to the max batch size before reporting the batch synchronously (the next process step does not execute until the batch is flushed). If a long-running step occurs (e.g. ansible), then a background timer will force a flush within the configured duration (default of 15 seconds). A flush also occurs when the process terminates (suspends, or final termination).


ibodrov (approved):
LGTM
jFYI, for stuff like ReportingContext we can use Java records now.

PR #954, opened 2024-08-08
concord-server: allow non-standard runtimes

ibodrov:
Add `process.extraRuntimes` configuration parameter to concord-server.conf. Only the default v1, v2 and `extraRuntimes` will be allowed as `configuration.runtime` values.

> what's the result if someone misspells the runtime? e.g. runtime: runtime-v@

The process will transition NEW -> FAILED with the exception printed in the "system" log segment.


brig (approved):

ibodrov:
Good point. I think I will make a config option `availableRuntimes` or something like that. We might be able to validate it even earlier than ProjectLoader.

ibodrov:
FTR, my plan is to gradually refactor all this to the state when we can define new runtimes in server plugins.

brig (approved):

PR #959, opened 2024-08-21
server: skip pull_request process start when useEventCommitId and event is from different repo

benbroadaway:
Processes will always fail if a github trigger enables `useEventCommitId: true` and an event related to a pull request (`pull_request`, `pull_request_review_comment`, `pull_request_review`, etc) is generated from a different repo (e.g. a fork). This will skip the waste of resources for attempting to start processes which always fail. The one downside is there's not much of an indication for end users.


server/impl/src/main/java/com/walmartlabs/concord/server/events/github/PullRequestPayload.java
```diff
@@ -0,0 +1,46 @@
+package com.walmartlabs.concord.server.events.github;
+
+import com.walmartlabs.concord.sdk.MapUtils;
+
+import java.util.Map;
+import java.util.Objects;
+
+public class PullRequestPayload extends Payload {
```
ibodrov, line 8: It looks like PullRequestPayload doesn't carry any extra "state" compared to Payload -- just some extra behaviour. Perhaps the same thing can be achieved with an utility function instead?

Something like:
```java
GithubTriggerV2Processor  {

  private boolean isPullRequestFromSameRepo(Payload payload) {
    // ...rest of the owl
  }
}
``` 

And we won't need `payload instanceof PullRequestPayload prp` anymore.

ibodrov: Basically, it looks like it can all be folded into `skipTrigger` function.

benbroadaway: I think that's fair at this level of complexity. The polymorphism is a bit obfuscated since we keep any variety of payload in a `Map`. If, for some reason, we expand functions considerably to <do stuff> with many different event types then it starts to make organizational sense to contain the event-specific methods to a dedicated class.

Also, upon some more thought, this specific change needs to consider more than just `pull_request` events. Really it's any event that may contain a `pull_request` _attribute_ in the payload. While it doesn't make a whole lot of sense to set `useEventCommitId: true` on for a `pull_request_review` event, it's _technically_ possible to attempt it in the trigger config.

ibodrov (approved):
LGTM

PR #977, opened 2024-09-18
runtime-v2: add github exclusive trigger to schema

benbroadaway:
`groupBy` currently isn't represented in the [runtime-v2 json schema](https://repo1.maven.org/maven2/com/walmartlabs/concord/runtime/v2/concord-runtime-model-v2/2.14.0/concord-runtime-model-v2-2.14.0-schema.json).

### DIFF
--- a/runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/ExclusiveMode.java
+++ b/runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/ExclusiveMode.java
@@ -20,22 +20,12 @@ package com.walmartlabs.concord.runtime.v2.model;
  * =====
  */
 
-import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import org.immutables.value.Value;
 
 import java.io.Serializable;
 
-@Value.Immutable
-@Value.Style(jdkOnly = true)
-@JsonInclude(JsonInclude.Include.NON_EMPTY)
-@JsonSerialize(as = ImmutableExclusiveMode.class)
-@JsonDeserialize(as = ImmutableExclusiveMode.class)
 public interface ExclusiveMode extends Serializable {
-
-    long serialVersionUID = 1L;
 
     @Value.Parameter
     @JsonProperty(value = "group", required = true)
@@ -64,7 +54,4 @@ public interface ExclusiveMode extends Serializable {
         wait
     }
 
-    static ExclusiveMode of(String group, Mode mode) {
-        return ImmutableExclusiveMode.of(group, mode);
-    }
 }
--- a/runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/DefaultExclusiveMode.java
+++ b/runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/DefaultExclusiveMode.java
@@ -4,7 +4,7 @@ package com.walmartlabs.concord.runtime.v2.model;
  * *****
  * Concord
  * -----
- * Copyright (C) 2017 - 2020 Walmart Inc.
+ * Copyright (C) 2017 - 2024 Walmart Inc.
  * -----
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


dankle (dismissed):

runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/ExclusiveMode.java
```diff
@@ -65,6 +60,19 @@ enum Mode {
...
-        return ImmutableExclusiveMode.of(group, mode);
+        return ImmutableDefaultExclusiveMode.of(group, mode);
+    }
+
+    @Value.Immutable
+    @Value.Style(jdkOnly = true)
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonSerialize(as = ImmutableDefaultExclusiveMode.class)
+    @JsonDeserialize(as = ImmutableDefaultExclusiveMode.class)
+    interface DefaultExclusiveMode extends ExclusiveMode {
```
ibodrov, line 71: Not sure why do we need a separate interface here?

benbroadaway: One `@Immutable` class should not extends another `@Immutable` class.
```java
@Value.Immutable
public interface MyInterface {
    boolean isTrue();
}

@Value.Immutable
public interface AnotherInterface extends MyInterface {
    boolean isFalse();
}
```

```
AnotherInterface.java: (immutables:subtype) Should not inherit com.example.MyInterface which is
a value type itself. Avoid extending from another abstract value type. Better to share common
abstract class or interface which are not carrying @Immutable annotation. If still extending from
immutable abstract type be ready to face some incoherences in generated types.
```

benbroadaway: > Does it change the serialization of TriggerMixIn?

Is the concern in this question regarding somewhere we deserialize it from a value stored in the db? i.e. a value stored pre-upgrade may fail after this change is in?

ibodrov (approved):

ibodrov:
@brig should we mark GithubTriggerExclusiveMode#group and #groupBy fields as @Deprecated?

runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/ExclusiveMode.java
ibodrov: I misunderstood what was happening here, my bad.

brig: I'm still not quite sure why a new class appeared here...

Wouldn't it be enough to just change:

GithubTriggerMixIn:
```
            @JsonProperty("exclusive")
            ExclusiveMode exclusive();
```
to
```
            @JsonProperty("exclusive")
            GithubTriggerExclusiveMode exclusive();
```

?

brig: something like this:
```
@@ -111,7 +112,7 @@ public interface TriggerMixIn extends Trigger {
             GithubTriggerConditions conditions();
 
             @JsonProperty("exclusive")
-            ExclusiveMode exclusive();
+            GithubTriggerExclusiveMode exclusive();
 
             interface GithubTriggerConditions {
                 @JsonProperty(value = "type", required = true)
@@ -174,6 +175,9 @@ public interface TriggerMixIn extends Trigger {
 
             @JsonProperty("conditions")
             Map<String, Object> conditions();
+
+            @JsonProperty("exclusive")
+            ExclusiveMode exclusive();
         }
     }
 
@@ -187,8 +191,5 @@ public interface TriggerMixIn extends Trigger {
 
         @JsonProperty("arguments")
         Map<String, Object> arguments();
-
-        @JsonProperty("exclusive")
-        ExclusiveMode exclusive();
     }
 }
```

ibodrov: Not sure if that would be reflected in the resulting schema?
Currently, it's 
```json
"triggers" : {
      "type" : "array",
      "items" : {
        "oneOf" : [ {
          "$ref" : "#/definitions/GenericTrigger"
        }, {
          "$ref" : "#/definitions/CronTrigger"
        }, {
          "$ref" : "#/definitions/ManualTrigger"
        }, {
          "$ref" : "#/definitions/GithubTrigger"
        }, {
          "$ref" : "#/definitions/OneOpsTrigger"
        } ]
      }
}
```
With GithubTrigger containing

```json
"GithubTriggerParams" : {
      "type" : "object",
      "additionalProperties" : false,
      "properties" : {
        "version" : {
          "type" : "integer",
          "enum" : [ 2 ]
        },
        "exclusive" : {
          "$ref" : "#/definitions/GithubTriggerExclusiveMode"
        },
        ...
```

and other trigger types using the base ExclusiveMode:

```json
"ManualTriggerParams" : {
      "type" : "object",
      "additionalProperties" : false,
      "properties" : {
        "name" : {
          "type" : "string"
        },
        "arguments" : {
          "type" : "object"
        },
        "activeProfiles" : {
          "type" : "array",
          "items" : {
            "type" : "string"
          }
        },
        "entryPoint" : {
          "type" : "string"
        },
        "exclusive" : {
          "$ref" : "#/definitions/ExclusiveMode"
        }
      },
      "required" : [ "entryPoint" ]
    },
```

Which seems fine to me?

ibodrov:
On a separate note, I don't think this is right?

```json
"ManualTrigger" : {
      "type" : "object",
      "additionalProperties" : false,
      "properties" : {
        "manual" : {
          "$ref" : "#/definitions/ManualTriggerParams"
        },
        "exclusive" : {
          "$ref" : "#/definitions/ExclusiveMode"
        }
      },
      "title" : "Manual Trigger"
    },
    "ManualTriggerParams" : {
      "type" : "object",
      "additionalProperties" : false,
      "properties" : {
        "name" : {
          "type" : "string"
        },
        "arguments" : {
          "type" : "object"
        },
        "activeProfiles" : {
          "type" : "array",
          "items" : {
            "type" : "string"
          }
        },
        "entryPoint" : {
          "type" : "string"
        },
        "exclusive" : {
          "$ref" : "#/definitions/ExclusiveMode"
        }
      },
      "required" : [ "entryPoint" ]
    },
```

Why both ManualTrigger and ManualTriggerParams have `exclusive` as a property? Doesn't look right to me. :eyes:

runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/ExclusiveMode.java
brig: > Not sure if that would be reflected in the resulting schema?
It should be, why not?

from master:
```
...
"GithubTriggerParams" : {
...
        "exclusive" : {
          "$ref" : "#/definitions/ExclusiveMode"
        },
....
```
withi just ExclusiveMode -> GithubTriggerExclusiveMode change:
```
...
"GithubTriggerParams" : {
...
        "exclusive" : {
          "$ref" : "#/definitions/GithubTriggerExclusiveMode"
        },
....

    "GithubTriggerExclusiveMode" : {
      "type" : "object",
      "additionalProperties" : false,
      "properties" : {
        "group" : {
          "type" : "string"
        },
        "groupBy" : {
          "type" : "string"
        },
        "mode" : {
          "type" : "string",
          "enum" : [ "cancel", "cancelOld", "wait" ]
        }
      }
    }
```

brig:
> On a separate note, I don't think this is right?

yep :(

[this](https://github.com/walmartlabs/concord/blob/master/runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/schema/TriggerMixIn.java#L48-L49) lines 

should be in `ManualTriggerParams`

brig (approved):

PR #981, opened 2024-09-19
runtime-v2: fix the issue where old agents can't parse the process configuration with new attributes

brig:


brig:
maybe we should put this annotation on all the other "models"?

ibodrov:
I think we should. Just to simplify the upgrades.

brig:
> I think we should. Just to simplify the upgrades.

done

ibodrov (approved):

benbroadaway (approved):

PR #1001, opened 2024-10-22
runtime-v2: flush events on process error

brig:
and `EventReportingService` does not implement the `ExecutionListener` interface, as it's not supposed to :)


ibodrov (approved):

brig:
We probably don't need the `EventReportingService` interface... What do you think?

ibodrov:
> We probably don't need the `EventReportingService` interface... What do you think?

Agree, let's remote it. I guess the original idea was to provide alternative implementations in, for example, the CLI -- i.e. print out events instead of sending to the server. But I see that in the CLI we just provide our own Set<ExecutionListeners>.

brig:
> > We probably don't need the `EventReportingService` interface... What do you think?
> 
> Agree, let's remote it. I guess the original idea was to provide alternative implementations in, for example, the CLI -- i.e. print out events instead of sending to the server. But I see that in the CLI we just provide our own Set.

will remove it in separate PR. after this one: https://github.com/walmartlabs/concord/pull/1002

PR #1014, opened 2024-11-04
runtime-v2: introduce extraDependencies

ibodrov:
Draft while I'm working on tests.

When active profiles are merged only one `dependencies` list "wins" --
currently is is only possible to override the dependency list entirely.

The new `configuration.extraDependencies` parameter provides a way to
add new dependencies using profiles. All values from all active profiles
are combined.

For example:

```yaml
configuration:
  runtime: "concord-v2"
  dependencies:
    - mvn://com.walmartlabs.concord.plugins.basic:ansible-tasks:2.15.0
profiles:
  foo:
    configuration:
      dependencies:
        - mvn://com.walmartlabs.concord.plugins.basic:ansible-tasks:2.16.0
  bar:
      extraDependencies:
        - mvn://com.walmartlabs.concord.plugins:confluence-task:2.5.0
flows:
  default:
    - log: "Hi!"
```

Running this YAML normally (i.e. `activeProfiles=default`) should result in a process with only `mvn://com.walmartlabs.concord.plugins.basic:ansible-tasks:2.15.0` as a dependency.
Running with `activeProfiles=foo` should replace the dependency with 2.16.0 version.
Selecting both profiles (`activeProfiles=foo,bar`) should result in both ansible-tasks:2.16.0 and confluence-task:2.5.0 being added.


cli/src/main/java/com/walmartlabs/concord/cli/Run.java
```diff
@@ -256,10 +270,11 @@ public Integer call() throws Exception {
                 new CliServicesModule(secretStoreDir, targetDir, defaultTaskVars, new VaultProvider(vaultDir, vaultId), dependencyManager, verbosity))
                 .create();
 
+        // TODO ask brig why [separate loader], why [same targetDir], why [no imports]
```
brig, line 273: At this point, the entire project is already loaded, all imports have been processed, and everything is placed in the `targetDir`. this load will handle all files, including the imported ones...

when we load the project for the first time, we don't yet have an Injector, no listeners will be triggered :(

I think these listeners aren’t actually used anywhere—probably best to just get rid of them:)

ibodrov: Agree https://github.com/walmartlabs/concord/pull/1015

brig (dismissed):

benbroadaway (dismissed):

brig (approved):

brig (approved):

PR #1027, opened 2024-11-07
mocks-task: record task events for mocked tasks

brig:
- Now mocked tasks will send events with in/out parameters. Everything will look as if the original task was called; 
- Now it's not possible to mock a method that doesn't exist in the original task. I think this is better to make the behavior closer to a flow without mocks;
- If a task is mocked, the original task won’t even be created to prevent it from executing anything in its constructor.


ibodrov (approved):

brig:
docs: https://github.com/walmartlabs/concord-website/pull/176

PR #1035, opened 2024-11-12
server: add authorization check to User lookup endpoint in API

laurayco:
restricts UserResourceV2 to only allow a user to look up their own information or requires admin permissions.

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserResourceV2.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/user/UserResourceV2.java
@@ -91,8 +91,8 @@ public class UserResourceV2 implements Resource {
 
         UUID authenticatedId = loggedIn.getId();
 
-        if(!authenticatedId.equals(id)) {
-            assertAdmin();
+        if(!authenticatedId.equals(id) && !Roles.isAdmin()) {
+            throw new UnauthorizedException("Users can only view their own information or must have admin privileges.");
         }
 
         return userDao.get(id);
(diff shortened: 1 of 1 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/user/UserResourceV2.java
```diff
@@ -66,12 +69,32 @@ public List<UserEntry> list(@QueryParam("offset") @DefaultValue("0") int offset,
...
     @Operation(description = "Get an existing user", operationId = "getUser")
     public UserEntry get(@PathParam("id") UUID id) {
+
+        UserPrincipal loggedIn = UserPrincipal.assertCurrent();
+
+        UUID authenticatedId = loggedIn.getId();
+
+        if(!authenticatedId.equals(id)) {
+            assertAdmin();
+        }
```
ibodrov, line 96: Let's inline the assertAdmin method and just do `...if (!Roles.isAdmin())...` right here. And let's be more specific in the error message. In the old UserResource the assert was strict all or nothing, and here it is more complicated.

```suggestion
        if(!authenticatedId.equals(id) && !Roles.isAdmin()) {
            throw new UnauthorizedException("Users can only view their own information or must have admin privileges.");
        }
```

or something like that.

ibodrov (changes requested):

server/impl/src/main/java/com/walmartlabs/concord/server/user/UserResourceV2.java
```diff
@@ -66,12 +69,32 @@ public List<UserEntry> list(@QueryParam("offset") @DefaultValue("0") int offset,
...
+    /*
+        throws unauthorized exception if user is not an admin,
+        copied from UserResource.java
+    */
+    private static void assertAdmin() {
+        if (!Roles.isAdmin()) {
+            throw new UnauthorizedException("Only admins can do that");
+        }
+    }
+
```
brig, line 82: unused, I think

```suggestion
```

ibodrov (approved):

PR #1048, opened 2024-12-07
concord-server: avoid security subject creation when it is not needed

ibodrov:


server/impl/src/main/java/com/walmartlabs/concord/server/security/SecurityUtils.java
```diff
@@ -44,31 +44,37 @@ public final class SecurityUtils {
...
-        Subject s = getSubject();
-        return s.isPermitted(permission);
+        Subject subject = getSubject();
+        if (subject == null) {
+            return false;
+        }
+        return subject.isPermitted(permission);
+    }
+
+    public static void bindSubject(Subject subject) {
```
brig, line 69: maybe add `unbindSubject()`?
bind here, unding in `ThreadContext` ...

PR #1059, opened 2024-12-31
runtime-v2: throw UserDefinedException from TaskResult.fail to tone down noise in logs

brig:
For example, in the case of TaskResult.fail("Docker process finished with exit code " + code), there’s no point in throwing a `RuntimeException` and printing the stack trace...

Values from `TaskResult` are now passed as `payload` in `UserDefinedException`


brig:
https://github.com/walmartlabs/concord/pull/1062

PR #1060, opened 2025-01-02
concord-server: replace synchronized with locks

ibodrov:
In preparation for virtual thread support.


server/impl/src/main/java/com/walmartlabs/concord/server/boot/BackgroundTasks.java
```diff
@@ -24,25 +24,34 @@
...
 
 public class BackgroundTasks {
 
     private final Set<BackgroundTask> tasks;
+    private final Lock controlMutex = new ReentrantLock();
 
     @Inject
     public BackgroundTasks(Set<BackgroundTask> tasks) {
-        this.tasks = tasks;
+        this.tasks = Set.copyOf(tasks);
```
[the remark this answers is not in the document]
ibodrov: Good point

brig (approved):

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/TrustingSslSocketFactory.java
```diff
@@ -39,10 +41,17 @@
  */
 public class TrustingSslSocketFactory extends SocketFactory {
 
+    private static final Lock mutex = new ReentrantLock();
+
     private final SSLSocketFactory delegate;
 
-    public static synchronized SocketFactory getDefault() {
-        return new TrustingSslSocketFactory();
+    public static SocketFactory getDefault() {
```
brig, line 48: any idea why this method was synchronized?

brig (approved):

server/impl/src/main/java/com/walmartlabs/concord/server/security/ldap/TrustingSslSocketFactory.java
ibodrov: > any idea why this method was synchronized?

I vaguely remember having issues with some SslSocketFactories that fail in constructors when created concurrently.  I don't think that we have any tests that use LDAP over SSL, I'll try to test it manually -- see if maybe we can remote the synchronization.

PR #1069, opened 2025-01-17
project: use GitHub-hosted ARM64 runners

brig:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/


ibodrov:
> Successful in 138m 

So looks like it is running on an equivalent of a half of a Raspberry Pi 5 :-)

brig:
> > Successful in 138m
> 
> So looks like it is running on an equivalent of a half of a Raspberry Pi 5 :-)

yeah. super-slow:(

PR #1076, opened 2025-02-05
runtime-v2: do not log stack trace for EL MethodNotFound exception and unify the error messages

brig:


ibodrov (approved):
Some nitpicking, otherwise LGTM

PR #1089, opened 2025-02-24
concord-agent: allow skipping repository state

brig:


agent/src/main/java/com/walmartlabs/concord/agent/cfg/GitConfiguration.java
```diff
@@ -50,6 +51,7 @@ public GitConfiguration(Config cfg) {
         this.httpLowSpeedTime = cfg.getDuration("git.httpLowSpeedTime");
         this.sshTimeout = cfg.getDuration("git.sshTimeout");
         this.sshTimeoutRetryCount = cfg.getInt("git.sshTimeoutRetryCount");
+        this.enabled = !cfg.hasPath("git.enabled") || cfg.getBoolean("git.enabled");
```
ibodrov, line 54: Add `git.enabled` to the default config and remove the hasPath check?

ibodrov (approved):

brig:
> Why might one be inclined to use this option?

I have N agents, some of which don’t have access to GH. For these agents, `CustomEnqueueProcessor.handleState` generates the process state, which is then saved in the database by `StateImportingProcessor`. Then state passed to the agent... The process appears as if it was triggered from GH :)

brig:
> Makes sense. So the tradeoff would be larger state payloads delivered to the agent, and some amount (maybe negligible, depends on how the custom processor works) slower time to get the process to `ENQUEUED` status.

Yeah, for Walmart, this is probably not relevant. 

We could have done this in some custom task (instead of the standard Concord one), but this way seemed simpler. Nothing changes for the user: they just call the Concord task, and how the payload gets to the agent is just some server-side magic...

PR #1092, opened 2025-02-27
tasks: new env and collections tasks

brig:


plugins/tasks/misc/src/main/java/com/walmartlabs/concord/plugins/misc/CollectionsTaskV2.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Creates a new list of integers from 0 to the specified size (exclusive).
+     * Each element in the list corresponds to its index.
+     *
+     * @param size the size of the list to create
+     * @return a list of integers from 0 to (size - 1)
+     */
+    public static List<Integer> newList(int size) {
```
ibodrov, line 65: Usually a function like this is called `range` :-)

ibodrov (approved):

plugins/tasks/misc/src/main/java/com/walmartlabs/concord/plugins/misc/CollectionsTaskV2.java
brig: > Usually a function like this is called `range` :-)

```suggestion
    public static List<Integer> range(int size) {
```

ibodrov (approved):

PR #1096, opened 2025-03-06
server: default message for unexpected errors

benbroadaway:
Don't let stack traces through for unexpected errors.

## Before

Screenshot 2025-03-06 at 8 06 28 AM

## After

Screenshot 2025-03-06 at 8 07 18 AM

### DIFF
--- a/plugins/tasks/misc/src/main/java/com/walmartlabs/concord/plugins/misc/CollectionsTaskV2.java
+++ b/plugins/tasks/misc/src/main/java/com/walmartlabs/concord/plugins/misc/CollectionsTaskV2.java
@@ -1,5 +1,25 @@
 package com.walmartlabs.concord.plugins.misc;
 
+/*-
+ * *****
+ * Concord
+ * -----
+ * Copyright (C) 2017 - 2025 Walmart Inc.
+ * -----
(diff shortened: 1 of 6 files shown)


ibodrov:
Can you add the stacktrace to the page? Maybe hidden under a "details" link. Might be useful for user support.

mtkunkel:
Perhaps a reference string on the page that matches the server logs would suffice? 


We have a request to stop dumping that uncontrolled output to the user....

ibodrov:
I thought about maybe a div that can be clicked on to see the actual error. "Closed" by default . But it's ok if there is a hard requirement to never even render the stacktrace.

ibodrov (approved):

benbroadaway:
Well, as it is, the stack trace isn't available within the context of a `RequestErrorHandler` instance. So it's not possible without some kind of refactoring...I think. Probably need some further consideration to force a stack track in the server logs if we want to guarantee it'll exist. IDK, don't want too much stack trace junk in the logs, but also don't wanna spew it straight to clients that are up to no good. Maaaaaybe a server config option to allow them through (e.g. flip it on temporarily for debug, or in a test env) in the case of spooky problem where the cause isn't easy to discern from the request?

PR #1100, opened 2025-03-12
console: handle non-existent root parent

benbroadaway:
Problem: UI crashes when a child process exists, but it's "root" parent process does not. For instance, if the parent starts the child a day (or more) after the parent was created, but the parent is cleaned up. Then the UI will crash when the root it searched for.


console2/src/components/organisms/ProcessActivity/index.tsx
```diff
@@ -131,7 +131,15 @@ const ProcessActivity = (props: ExternalProps) => {
         setProcess(process);
 
         if (process.parentInstanceId && !rootProcessRef.current) {
-            rootProcessRef.current = await apiGetRoot(process.parentInstanceId);
+            apiGetRoot(process.parentInstanceId)
+                .catch(() => {
+                    rootProcessRef.current = null;
```
benbroadaway, line 136: Is it safe to set this to `process` (the current process) to avoid re-doing, and failing, this lookup each refresh?

benbroadaway: I don't _think_ so....since this `rootProcessRef` is used for process restart functionality

Screenshot 2025-03-12 at 8 15 06 AM

ibodrov: Maybe? It will be the same process instance anyway. Should not matter when we load it.

console2/src/components/organisms/ProcessActivity/index.tsx
```diff
@@ -131,7 +131,15 @@ const ProcessActivity = (props: ExternalProps) => {
...
+                })
+                .then((rootProc) => {
```
ibodrov, line 138: Why not use `await` with try-catch? The outer function is already async.

console2/src/api/process/index.ts
```diff
@@ -265,10 +265,15 @@ export const get = (
...
 };
 
-export const getRoot = (
-    instanceId: ConcordId
-): Promise<ProcessEntry> => {
-    return fetchJson(`/api/v1/process/${instanceId}/root`);
+export const getRoot = async (instanceId: ConcordId): Promise<ProcessEntry|void> => {
+    const resp = await managedFetch(`/api/v1/process/${instanceId}/root`, { method: 'GET' });
+    if (resp.status === 200) {
+        console.log(resp);
```
ibodrov, line 271: Debug logging? You should see the response in the dev tools anyway.

console2/src/components/organisms/ProcessActivity/index.tsx
benbroadaway: Fair. reverted to the simpler approach. Going in i thought I might have to account for more outcomes. Sometimes it's a 204 no-content response, others returns 400 bad request depending on parent-child(-child...) depth and what does or doesn't exist.

console2/src/api/process/index.ts
benbroadaway: indeed, wasn't intended to stick around

ibodrov (approved):

PR #1106, opened 2025-03-17
concord-server, concord-server-db: generate UUIDs on the server, use UUID v7

ibodrov:
UUIDv7 should be more b-tree friendly.
Assuming there are no INSERTs done directly into the DB. If there are direct INSERTs then we can leave the default values in place and still generate UUIDv7 in the server.


benbroadaway:
For the sake of backwards compatibility between `2.25.0` and `2.24.0`, could we leave the defaults for one version then remove them in `2.26.0`?

ibodrov:
Good point. I'll remove the db migration for now.

brig (approved):

PR #1108, opened 2025-03-18
concord-server, concord-client2: uncomment valid code

ibodrov:
Not sure why it was commented in the first place (leave comments next time, please :smile:). Let's see if it breaks anything.


benbroadaway:
re: the gauges

I'm not 100% sure why. But it was in a big PR for `inventory -> storage + UI` (merged as `concord-server: rework Concord Inventory as JSON Store API, make it GA `) in a commit named `tests up`. Not sure if it was flaky or something, or just an oopsie.

ibodrov:
I don't think we even test those gauges. Tests seems to be fine.

brig (approved):

PR #1121, opened 2025-03-28
project: use eatmydata to speed up builds

ibodrov:


ibodrov:
Didn't speed up anything.

PR #1122, opened 2025-03-28
runtime-v2: set lowercase loop mode value in schema

benbroadaway:


brig (approved):

brig:
I hope Uppercase also parsed by runtime-v2:
```
optional("mode", enumVal(Loop.Mode.class, String::equalsIgnoreCase).map(o::mode))
```

benbroadaway:
> I hope Uppercase also parsed by runtime-v2:
> 
> ```
> optional("mode", enumVal(Loop.Mode.class, String::equalsIgnoreCase).map(o::mode))
> ```

Yeah both work. We usually document values such as this in their lowercase form (e.g. [`mode: parallel`](https://concord.walmartlabs.com/docs/processes-v2/flows.html#loops). If the source is an enum, then the default options generated in the JSON Schema come out uppercase.

PR #1128, opened 2025-04-24
server: handle invalid regex in dispatcher requirements

benbroadaway:
A bad regex in a process request's agent requirements can completely halt the queue dispatcher since one exception causes the entire run to fail.

This change handles an exception per matching attempt so the dispatcher can carry on if any requests are catastrophically invalid. The process(es) with non-processable requirements are `FAILED`.

Screenshot 2025-04-24 at 12 07 14 PM

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/process/queue/dispatcher/Dispatcher.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/process/queue/dispatcher/Dispatcher.java
@@ -175,7 +175,16 @@ public class Dispatcher extends PeriodicTask {
             // filter out the candidates that shouldn't be dispatched at the moment (e.g. due to concurrency limits)
             for (ProcessQueueEntry e : candidates) {
                 // find request/agent who can handle process
-                Request req = findRequest(e, inbox);
+                Request req = findRequest(e, inbox, (queueEntry, ex) -> {
+                    if (ex instanceof PatternSyntaxException) {
+                        logManager.error(queueEntry.key(), "Invalid regex in requested agent capabilities.");
+                    } else {
+                        logManager.error(queueEntry.key(), "Error matching process request requirements.");
+                    }
+
+                    queueManager.updateStatus(tx, queueEntry.key(), ProcessStatus.FAILED);
+                });
+
                 if (req == null) {
                     continue;
                 }
@@ -213,14 +222,14 @@ public class Dispatcher extends PeriodicTask {
         return matches;
     }
 
-    static Request findRequest(ProcessQueueEntry candidate, List<Request> requests) {
+    static Request findRequest(ProcessQueueEntry candidate, List<Request> requests, RequirementsMatcherErrorHandler errHandler) {
         return requests.stream()
-                .filter(req -> isRequestMatch(candidate, req))
+                .filter(req -> isRequestMatch(candidate, req, errHandler))
                 .findFirst()
                 .orElse(null);
     }
 
-    private static boolean isRequestMatch(ProcessQueueEntry candidate, Request req) {
+    private static boolean isRequestMatch(ProcessQueueEntry candidate, Request req, RequirementsMatcherErrorHandler errHandler) {
         try {
             var capabilities = req.request.getCapabilities();
             var requirements = getAgentRequirements(candidate);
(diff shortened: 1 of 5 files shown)


brig (dismissed):

server/impl/src/main/java/com/walmartlabs/concord/server/process/queue/dispatcher/Dispatcher.java
```diff
@@ -211,28 +213,37 @@ private List<Match> match(DSLContext tx, List<Request> requests) {
...
+    }
+
+    private static boolean isRequestMatch(ProcessQueueEntry candidate, Request req) {
+        try {
+            var capabilities = req.request.getCapabilities();
+            var requirements = getAgentRequirements(candidate);
+
+            return requirements.isEmpty() || Matcher.matches(capabilities, requirements);
+        } catch (PatternSyntaxException pse) {
+            log.error("Invalid regex in requested agent capabilities for instanceId: {}", candidate.key().getInstanceId());
```
ibodrov, line 230: Maybe we should have a "shut up" command that we can send to agents with bad configuration. :-)
I.e. force those agents into maintenance mode.

benbroadaway: It's the reverse. Agent configs are fine in this case. However, If a `concord.yml` has a bad `configuration.requirements.agent.whatever` regex pattern, then it causes the dispatcher to explode and fail to dispatch any more processes.

benbroadaway: I do think some more defense against letting garbage in is better, but it's a little more of an open landscape to pre-validate (compile) all fields as a regular expression.

benbroadaway: I think further refinements here could be in the direction of kicking the nasty process request out so it won't continually be re-processed. Something like one of:
* Set the process to `FAILED` or `CANCELLED`. Fix your junk and try again.
* Assume no requirements if the provided value(s) cannot be read. You get dispatched to a default agent. *May* have side effects.

ibodrov: I feel like marking as FAILED is an appropriate reaction. With some amount of cursing in the process log :-)

server/impl/src/main/java/com/walmartlabs/concord/server/process/queue/dispatcher/Dispatcher.java
```diff
@@ -213,23 +222,26 @@ private List<Match> match(DSLContext tx, List<Request> requests) {
...
             var capabilities = req.request.getCapabilities();
             var requirements = getAgentRequirements(candidate);
 
             return requirements.isEmpty() || Matcher.matches(capabilities, requirements);
         } catch (PatternSyntaxException pse) {
             log.error("Invalid regex in requested agent capabilities for instanceId: {}", candidate.key().getInstanceId());
+            errHandler.handle(candidate, pse);
         } catch (Exception e) {
-            log.error("Error matching process requirements: {}", e.getMessage());
+            var errMsg = String.format("Error matching process requirements for instanceId: %s", candidate.key().getInstanceId());
```
ibodrov, line 242: I propose we don't mix var/no-var within one .java file.

server/impl/src/main/java/com/walmartlabs/concord/server/process/queue/dispatcher/Dispatcher.java
```diff
@@ -175,7 +175,16 @@ private List<Match> match(DSLContext tx, List<Request> requests) {
             // filter out the candidates that shouldn't be dispatched at the moment (e.g. due to concurrency limits)
             for (ProcessQueueEntry e : candidates) {
                 // find request/agent who can handle process
-                Request req = findRequest(e, inbox);
+                Request req = findRequest(e, inbox, (queueEntry, ex) -> {
```
ibodrov, line 178: Make the error handler a class member instead? I.e. init it once in the constructor?

ibodrov (commented):
Minor comments

server/impl/src/main/java/com/walmartlabs/concord/server/process/queue/dispatcher/Dispatcher.java
benbroadaway: That's fair 👍

ibodrov (approved):

PR #1139, opened 2025-05-06
Default Git Credentials Hostname Whitelist

laurayco:
Adds a field to the git client configuration which serves as a whitelist for allowed hosts to use the default git credentials with (to prevent leaking of said credentials.) WIP Label bc I'm not confident that I've covered all bases and would appreciate more eyes on it :)

### DIFF
--- a/repository/src/main/java/com/walmartlabs/concord/repository/GitClient.java
+++ b/repository/src/main/java/com/walmartlabs/concord/repository/GitClient.java
@@ -302,20 +302,11 @@ public class GitClient {
     private String updateUrl(String url, Secret secret) {
         url = url.trim();
 
-        int hostnameStart = url.indexOf("://") + 3;
-        int hostnameEnd = url.indexOf("/", hostnameStart);
-
-        if(hostnameStart==2 || hostnameEnd < 0) {
-            // probably this is invalid, but I don't know what proper handling of this looks like
-            // however you definitely don't want to give credentials to whatever this git url looks like.
-            return url;
-        }
-
-        String hostname = url.substring(hostnameStart, hostnameEnd);
+        URI uri = URI.create(url);
 
         List<String> allowedDefaultHosts = cfg.authorizedGitHosts();
 
-        if(secret == null && allowedDefaultHosts!= null && !allowedDefaultHosts.contains(hostname)) {
+        if(secret == null && allowedDefaultHosts!= null && !allowedDefaultHosts.contains(uri.getHost())) {
             // in this case the user has not provided authentication AND the host is not in the whitelist of hosts
             // which may use the default git credentials. return the url un-modified to attempt anonymous auth;
             // if it fails
--- a/repository/src/main/java/com/walmartlabs/concord/repository/GitClient.java
+++ b/repository/src/main/java/com/walmartlabs/concord/repository/GitClient.java
@@ -302,10 +302,35 @@ public class GitClient {
     private String updateUrl(String url, Secret secret) {
         url = url.trim();
 
-        if (secret != null || cfg.oauthToken() == null || url.contains("@") || !url.startsWith("https://")) {
+        int hostnameStart = url.indexOf("://") + 3;
+        int hostnameEnd = url.indexOf("/", hostnameStart);
+
+        if(hostnameStart==2 || hostnameEnd < 0) {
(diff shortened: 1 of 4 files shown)


ibodrov:
I thought the SSH protocol already prevents "private key leaks" if the host doesn't have the public part of the key authorized?

repository/src/main/java/com/walmartlabs/concord/repository/GitClient.java
```diff
@@ -302,10 +302,35 @@ private static String getRefSpec(Ref ref) {
...
+        int hostnameStart = url.indexOf("://") + 3;
+        int hostnameEnd = url.indexOf("/", hostnameStart);
+
+        if(hostnameStart==2 || hostnameEnd < 0) {
+            // probably this is invalid, but I don't know what proper handling of this looks like
+            // however you definitely don't want to give credentials to whatever this git url looks like.
+            return url;
+        }
+
+        String hostname = url.substring(hostnameStart, hostnameEnd);
```
ibodrov, line 314: All above can be replaced with `new URI(url)`, I think.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URI.html#%3Cinit%3E(java.lang.String)

repository/src/main/java/com/walmartlabs/concord/repository/GitClient.java
```diff
@@ -302,10 +302,35 @@ private static String getRefSpec(Ref ref) {
...
+        }
+
         if (secret != null || cfg.oauthToken() == null || url.contains("@") || !url.startsWith("https://")) {
+            // provided url already has credentials OR there are no default credentials to use.
+            // anonymous auth is the only viable option.
             return url;
         }
 
+        // using default credentials
+        System.out.println("Using default credentials: " + "https://" + cfg.oauthToken() + "@" + url.substring("https://".length()));
```
ibodrov, line 333: Let's not log the token. `GitClient` is used in several places, not just on the server side.

ibodrov: Tbh, I'd rather not to log anything here.

repository/src/main/java/com/walmartlabs/concord/repository/GitClientConfiguration.java
```diff
@@ -24,13 +24,16 @@
...
 import java.time.Duration;
+import java.util.List;
 
 @Value.Immutable
 @Value.Style(jdkOnly = true)
 public interface GitClientConfiguration {
 
     @Nullable
     String oauthToken();
+    @Nullable
```
ibodrov, line 35: Missing newline between like 34 and 35.

ibodrov (changes requested):

repository/src/main/java/com/walmartlabs/concord/repository/GitClient.java
laurayco: This was a discussion with @benbroadaway - I opted to avoid using URI because I wanted to avoid creating unnecessary objects for GC (my first impl was using URI); Java isn't my area of expertise, I'm happy to do whichever method is accepted with consensus :)

laurayco:
> I thought the SSH protocol already prevents "private key leaks" if the host doesn't have the public part of the key authorized? Is this whitelist for http(s) connections?

Well, SSH doesn't transmit the private key it just proves it has it according to the public key so even in that case it's fine. That being said, this is for the HTTPS schema which uses a token authentication that is susceptible to reuse. It was a finding from a security review I believe.

repository/src/main/java/com/walmartlabs/concord/repository/GitClient.java
ibodrov: I would not worry about GC pressure in this case. This method is not on a critical path in some tight loop or anything.

benbroadaway: Right, the discussion was on that general idea of when to be extra considerate of object creation and to double-check. The calls that came to mind which will run through this code (repo refreshes, process start requests) put it in the low-activity category (e.g. ~dozens per minute).

laurayco (commented):
all issues should be resolved now

ibodrov (approved):

PR #1142, opened 2025-05-16
crypto-tasks: add option to create secrets from string values

ibodrov:
```yaml
    - task: "crypto"
      in:
        action: create
        org: "Default"
        secretName: "mySecret"
        stringValue: "foobar"
```


benbroadaway:
Is there an advantage over the existing [`secrets`](https://github.com/walmartlabs/concord/blob/master/plugins/tasks/concord/src/main/java/com/walmartlabs/concord/client/v2/SecretsTaskV2.java) tasks?

ibodrov:
> Is there an advantage over the existing [`secrets`](https://github.com/walmartlabs/concord/blob/master/plugins/tasks/concord/src/main/java/com/walmartlabs/concord/client/v2/SecretsTaskV2.java) tasks?

I forgot about SecretsTaskV2 :skull: 
We should probably print a deprecation message in CryptoTaskV2#execute, with the recommendation to use the `secrets` task.

PR #1143, opened 2025-05-16
concord-cli: support for remote secrets, configurable secrets providers

ibodrov:
Config example:

```yaml
# ~/.concord/cli.yaml
contexts:
  default: # those are the defaults, we don't need to put them in the local config file
    secrets:
      vault:
        dir: "${configDir}/vaults"
        id: "default"
      local:
        enabled: true
        writable: true
        dir: "${configDir}/secrets"
      remote:
        enabled: false
        writable: false
        confirmAccess: true
  prod: # alternative "context", activated with "concord-cli run --context prod ..."
    secrets:
      local:
        enabled: false        
      remote:
        enabled: true
        baseUrl: "http://concord.prod.acme.com"
        apiKey: "foobar"
```

In future, the feature can be extended to support multiple providers of the same type.


cli/src/main/java/com/walmartlabs/concord/cli/runner/secrets/RemoteSecretsProvider.java
```diff
@@ -0,0 +1,165 @@
...
+public class RemoteSecretsProvider implements SecretsProvider {
+
+    private final Path workDir;
+    private final SecretClient secretClient;
+
+    public RemoteSecretsProvider(Path workDir, String baseUrl, String apiKey) {
+        this.workDir = requireNonNull(workDir);
+
+        var apiClient = new DefaultApiClientFactory(requireNonNull(baseUrl))
+                .create(ApiClientConfiguration.builder()
```
benbroadaway, line 45: I think it may be useful to set a user agent on the client here to identify calls from the CLI in a real environment.

```java
        var apiClient = new DefaultApiClientFactory(requireNonNull(baseUrl))
                .create(ApiClientConfiguration.builder()
                        .apiKey(requireNonNull(apiKey))
                        .build())
                .setUserAgent("Concord-CLI");
```

cli/src/main/java/com/walmartlabs/concord/cli/runner/secrets/CliSecretService.java
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+        throw new RuntimeException("Secret not found: %s/%s.".formatted(orgName, secretName));
+    }
+
+    @Override
+    public UsernamePassword exportCredentials(String orgName, String secretName, String secretPassword) throws Exception {
+        for (var ref : secretsProviders) {
+            var result = ref.provider().exportCredentials(orgName, secretName, secretPassword);
+            if (result.isPresent()) {
```
benbroadaway, line 106: Giving this a shot trying to look up a remote secret (username/password). Config loads fine, but the first attempt through this loop (with two providers available) ends with an exception so the second attempt (remote) never occurs.

Disabling `secrets.local` fixes that since it only needs to try the one, correct, provider. But it appears this logic is supposed to "find the first successfully loadable secret".

benbroadaway (changes requested):

cli/src/main/java/com/walmartlabs/concord/cli/runner/secrets/CliSecretService.java
```diff
@@ -0,0 +1,165 @@
...
+        if (local.enabled()) {
+            var provider = new FileSecretsProvider(workDir, local.dir());
+            providers.add(new SecretsProviderRef("localFile", provider, local.writable()));
+        }
+
+        var remote = cliConfig.secrets().remote();
+        if (remote.enabled()) {
+            var provider = new RemoteSecretsProvider(workDir, remote.baseUrl(), remote.apiKey());
+            providers.add(new SecretsProviderRef("remote", provider, remote.writable()));
+        }
```
benbroadaway, line 54: I'm a littttle bit concerned this opens a door to a big potential oopsie. One should be very considerate of their handling of secrets from "real" environments to start with, but this could easily end up with a "I got out of a pickle with the CLI pointed at prod real quick on Monday and forgot to turn off remote secrets...Tuesday got _interesting_  😬"

That could be addressed with an extra setting for the remote config(s), or any config since the same situation _is possible_--just fewer hoops to jump through (e.g. creating specific secrets locally for `secrets.local` vs `secrets.remote` unlocking any available secret in the system). Something like `secrets.remote.warnOnAccess: true` that gives a prompt/warning that it's about to access a remote secret.

ibodrov: That is a valid point. I will see if we can add a prompt there.
Or maybe providers should be configured for specific "contexts" and the CLI can require something like `--context prod" to use a non-default context?

cli/src/main/java/com/walmartlabs/concord/cli/runner/secrets/CliSecretService.java
ibodrov: Let me re-check this. It was supposed to try each provider until the secret is found.

cli/src/main/java/com/walmartlabs/concord/cli/runner/secrets/RemoteSecretsProvider.java
ibodrov: Good idea, will do.

cli/src/main/java/com/walmartlabs/concord/cli/runner/secrets/CliSecretService.java
ibodrov: Should be fixed now.

cli/src/main/java/com/walmartlabs/concord/cli/runner/secrets/CliSecretService.java
ibodrov: @benbroadaway how about now -- see the example in the PR description.
Users will need to explicitly say `concord run --context myNonDefaultContext`.

ibodrov: Added the confirmation prompt as well

image

brig (approved):

PR #1147, opened 2025-05-22
runtime-v1, runtime-v2: refactor loader structure

ibodrov:
- split ConcordProjectLoader into separate classes in the v1 and v2 runtimes.
- move common model interfaces from concord-process-loader into a separate module.
- create a `ProjectLoader` that delegates to the new v1 and v2 loaders.
- rename modules for consistency.

Now the module dependencies should look like this:

```
concord-runtime-model <--- concord-runtime-loader <--- concord-runtime-model-v1 <----+
                                    ^          ^                                     |
                                    |          |                                     |
                                    |          +----- concord-runtime-model-v2  <--- concord-server-impl
                                    |                                                |
                                    +------------------------------------------------+
```

This change enables us to add new runtimes by simply supplying a new `ProjectLoader`, possibly from an external dependency.


runtime/model/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        <!-- Immutables -->
+        <dependency>
+            <groupId>org.immutables</groupId>
+            <artifactId>value</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.immutables</groupId>
+            <artifactId>serial</artifactId>
+        </dependency>
```
benbroadaway, line 32: Is `org.immutables:serial` necessary here? I think it's just used by `imports` which should pull it into the transitive deps, right?

ibodrov: There's stuff like com.walmartlabs.concord.runtime.model.FormField that uses immutables.

brig (approved):

PR #1148, opened 2025-05-22
project: build with postgres 14 image

benbroadaway:


it/console/pom.xml
```diff
@@ -15,7 +15,7 @@
 
     <properties>
         <!-- pin down the version of DB and Selenium images -->
-        <db.image>library/postgres:10.4-alpine</db.image>
+        <db.image>library/postgres:14.18-alpine</db.image>
```
ibodrov, line 18: Maybe we should use this chance and move db.image to the parent pom?

brig (approved):

PR #1151, opened 2025-05-23
runtime-v2: allow marking nested values as sensitive data

brig:


runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/el/resolvers/SensitiveDataProcessor.java
```diff
@@ -53,11 +56,41 @@ public static void process(Object value, Method method) {
...
+        } else if (obj instanceof List<?> list) {
+            for (Object o : list) {
+                processNestedValues(o, depth + 1);
+            }
+        } else if (obj instanceof Set<?> set) {
+            for (Object item : set) {
+                processNestedValues(item, depth + 1);
+            }
+        } else if (obj != null && obj.getClass().isArray()) {
+            var len = Array.getLength(obj);
```
ibodrov, line 84: No mixing var/no-var in the same file, please.

ibodrov (changes requested):
Some nitpicking, otherwise LGTM

runtime/v2/runner/src/main/java/com/walmartlabs/concord/runtime/v2/runner/el/resolvers/SensitiveDataProcessor.java
brig: fixed!

ibodrov (approved):

PR #1160, opened 2025-06-04
project: use central-publishing-maven-plugin

ibodrov:
I have no idea why the project `<name>` is not being inherited from the parent POM. Worked for concord-plugins. :man_shrugging: 

Also, the artifact produced by really-executable-jar-maven-plugin is being rejected by the central portal for whatever reason. Disabling it for now, will work on a workaround/replacement in a patch release.

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -13,6 +13,8 @@
     <artifactId>concord-cli</artifactId>
     <packaging>jar</packaging>
 
+    <name>${project.groupId}:${project.artifactId}</name>
+
     <properties>
         <main.class>com.walmartlabs.concord.cli.Main</main.class>
     </properties>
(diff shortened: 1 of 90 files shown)


cli/pom.xml
```diff
@@ -192,25 +194,6 @@
...
-                        </goals>
-                    </execution>
-                </executions>
-                <configuration>
-                    <flags>-Xmx128m -client</flags>
-                    <classifier>uber</classifier>
-                    <outputClassifier>executable</outputClassifier>
-                    <allowOtherTypes>true</allowOtherTypes>
-                </configuration>
-            </plugin>
```
benbroadaway, line 213: Any particular reason this is removed?

ibodrov: > Also, the artifact produced by really-executable-jar-maven-plugin is being rejected by the central portal for whatever reason. Disabling it for now, will work on a workaround/replacement in a patch release.

ibodrov: Something-something, unknown artifact type.

benbroadaway: ah dang, jumped into the changes first.

benbroadaway (approved):

PR #1161, opened 2025-06-04
slack: allow send blocks instead of simple text

brig:


plugins/tasks/slack/src/main/java/com/walmartlabs/concord/plugins/slack/SlackTask.java
```diff
@@ -56,7 +56,7 @@ public void call(@InjectVariable("context") Context ctx,
                      boolean ignoreErrors) {
 
         SlackConfiguration slackCfg = SlackConfiguration.from(SlackConfigurationParams.of(new ContextVariables(ctx), defaults(ctx)));
-        delegate.sendMessage(slackCfg, channelId, ts, false, text, iconEmoji, username, attachments, ignoreErrors);
+        delegate.sendMessage(slackCfg, channelId, ts, false, text, iconEmoji, username, attachments, null, ignoreErrors);
```
benbroadaway, line 59: Nice 👍  I was thinking we should omit this support from runtime-v1 for two reasons:
1. Incentive to migrate to runtime-v2
2. `blocks` is a simple enough parameter/variable name that it _may_ result in breakage if it's floating around in existing flows.

PR #1165, opened 2025-06-10
add a whitelist for repo URI schemes to prevent abuse.

laurayco:


github-app-installation/src/main/java/com/walmartlabs/concord/github/appinstallation/CacheKey.java
```diff
@@ -0,0 +1,66 @@
...
+import org.immutables.value.Value;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.net.URI;
+import java.util.Objects;
+
+@Value.Immutable
+@Value.Style(jdkOnly = true, redactedMask = "**redacted**")
+interface CacheKey {
```
ibodrov:
```suggestion
record CacheKey(URI repoUri, int weight)
```

maybe? Calculate the weight when the key is created. Get rid of immutables and there won't be anything secret to print in the first place.

PR #1170, opened 2025-06-12
runtime-v2: support expressions for output variables in call step

brig:
something like this:

```
flows:
  default:
    - call: flowWrapper
      in:
        flow: "myFlow"
        flowArgs:
          key: value
        flowOut:
          - "myFlowOut"
      out:
        - "myFlowOut"

    - log: "${myFlowOut}"

  ##
  #  in:
  #    flow: string, mandatory, flow to execute
  #    flowArgs: object, mandatory, flow args
  #    flowOut: string[], mandatory, flow out variables
  ##
  flowWrapper:
    - log: "Start"

    - call: "${flow}"
      in: "${flowArgs}"
      out: "${flowOut}"

    - log: "End"

  ##
  #  out:
  #    myFlowOut: string, mandatory
  ##
  myFlow:
    - set:
        myFlowOut: "abc"
```

### DIFF
--- a/runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/FlowCallOptions.java
+++ b/runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/FlowCallOptions.java
@@ -59,7 +59,7 @@ public interface FlowCallOptions extends StepOptions {
         return Collections.emptyMap();
     }
 
-    @Deprecated(forRemoval = true, since = "2.28.0")
+    @Deprecated(forRemoval = true, since = "2.29.0")
     @Value.Default
     @AllowNulls
     default Map<String, Serializable> outExpr() {


runtime/v2/model/src/main/java/com/walmartlabs/concord/runtime/v2/model/FlowCallOptions.java
```diff
@@ -50,6 +50,16 @@ default List<String> out() {
...
+    @Nullable
+    String outExpression();
+
+    @Value.Default
+    @AllowNulls
+    default Map<String, Serializable> outMapping() {
+        return Collections.emptyMap();
+    }
+
+    @Deprecated(forRemoval = true, since = "2.28.0")
```
ibodrov, line 62:
```suggestion
    @Deprecated(forRemoval = true, since = "2.29.0")
```

ibodrov (approved):
Minor comment, otherwise LGTM

PR #1176, opened 2025-06-29
concord-server: config option to disable template script processing

benbroadaway:
Adds a kill switch for any template script execution on the server. Needs some further consideration for sandboxing the scripts when enabled. Newer graalvm SDKs may offer some extra options, but SecurityManager isn't really feasible (and is deprecated).


benbroadaway:
> maybe just remove this script processor? :)

That may be the ultimate outcome. It's a feature that appears really cool on paper, but it's too loosey-goosey in its current state. This is a middle-ground with a quick backout in case someone's using it and slipped through pre-upgrade scanning.

ibodrov (approved):

dankle (approved):

PR #1182, opened 2025-07-02
concord-agent: configurable list of runners

ibodrov:
Refactor agent configuration of v1 and v2 "runners". Now it is possible to add additional "runners" in an external config:
```
concord-agent {
    runnerV3 = ${runner} # reference to the existing common block of options
    runnerV3 {
        path = "/path/to/runner.jar"
        mainClass = "com.acme.runtime.v3.runner.Main"
        segmentedLogs = true
    }

    runtimes = {
       "concord-v3" = ${runnerV3}
        # will be merged with "concord-v1" and "concord-v2" keys
    }
}
```


brig:
Do we want anything regarding backward compatibility?

ibodrov:
Removed the nested blocks in favor of simpler backward-compatibility with existing concord-agent.confs.

benbroadaway (approved):

PR #1204, opened 2025-08-11
resource-task: Use the shared IOUtils assertInPath method for resource lookups

laurayco:
Using the IOUtils assertInPath to reduce duplicate code.

### DIFF
--- a/plugins/tasks/resource/pom.xml
+++ b/plugins/tasks/resource/pom.xml
@@ -26,6 +26,16 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>com.walmartlabs.concord</groupId>
+            <artifactId>concord-common</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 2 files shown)


plugins/tasks/resource/pom.xml
```diff
@@ -25,6 +25,16 @@
...
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.walmartlabs.concord</groupId>
+            <artifactId>concord-common</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
```
ibodrov, line 35: Why does it need commons-compress now? IIRC, it only needed if you use `zip*` methods from concord-common.

benbroadaway: `commons-compress` is not needed by the resource task, but it _is_ a transitive dependency of `concord-common`. Since `concord-common` is `provided` scope, its transitive deps aren't loaded for the unit tests. It could probably just be `test` scope instead of provided here.

ibodrov (approved):

PR #1209, opened 2025-08-27
docs: move core docs into main repository

ibodrov:
Copy core documentation from concord-website into the main repository. Use mdBook for building.


docs/book.toml
```diff
@@ -0,0 +1,15 @@
+[book]
+authors = ["Concord Authors"]
+language = "en"
+src = "src"
+title = "Concord"
+
+[preprocessor.variables.variables.site]
+concord_core_version = "2.32.0"
```
ibodrov, line 8: Can be replaced with `${project.version}` value, if we wire up the mdbook build through Maven.

PR #1309, opened 2026-04-23
concord-server: assert permissions for process details

ibodrov:
I think we can be stricter here.

Note: this can break existing workflows that rely on being able to fetch process details as an arbitrary Concord user.

### DIFF
--- a/it/server/src/test/java/com/walmartlabs/concord/it/server/RunAsIT.java
+++ b/it/server/src/test/java/com/walmartlabs/concord/it/server/RunAsIT.java
@@ -247,12 +247,13 @@ public class RunAsIT extends AbstractServerIT {
         Map<String, Object> data = Collections.singletonMap("msg", "Hello!");
         formsApi.submitForm(pe.getInstanceId(), formName, data);
 
-        // wait for the process to finish
+        // User B can submit the runAs form, but the process has no project and is still owned by user A.
+        setApiKey(apiKeyA.getKey());
 
+        // wait for the process to finish
         pe = waitForCompletion(getApiClient(), spr.getInstanceId());
 
         // check the logs
-
         resetApiKey();
 
         ab = getLog(pe.getInstanceId());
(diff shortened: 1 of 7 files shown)


Mrsandeep27 (commented):
Centralizing the process-access policy into \`ProcessAccessManager\` is a clean cleanup — better than the previous per-resource duplication. A few observations:

**1. \`assertAccess\` returns the entry on success.** That's a nice touch (avoids a second \`processQueueManager.get\` call by the caller), but be careful: callers that ignore the return value (\`accessManager.assertAccess(id, ...)\` as a statement) will silently swallow that work. If \`get\` is expensive enough that it warrants caching here, consider passing through a \`Loader\` callback so callers that already have an entry don't pay for re-fetching.

**2. The IT test change** — \`setApiKey(apiKeyA.getKey())\` is added with a comment explaining \"User B can submit the runAs form, but the process has no project and is still owned by user A.\" That's a behavior nuance worth elevating to a Javadoc on \`assertAccess\` itself, not just a test comment. The next contributor reading \`assertAccess\` from a different IDE pane won't see the IT comment.

**3. \`current.getId().equals(entry.initiatorId())\` short-circuits before the session-key check.** That's fine, but a process initiated by user A and being introspected by user B who happens to hold a session key delegated for that process: does the check work? The structure (initiator → session-key) is right but worth a positive test case for the session-key path, not just the initiator path.

**4. \`UnauthorizedException\` vs \`ConcordApplicationException(NOT_FOUND)\`.** Returning \`NOT_FOUND\` for missing processes and (presumably) \`UnauthorizedException\` for access-denied is the right shape (don't leak existence). The diff cuts off — confirm the \`else\` branch throws an unauthorized error rather than \`null\`/\`NOT_FOUND\`. If both branches return \`NOT_FOUND\`, that's a deliberate hide-existence behavior worth documenting; if one is auth and one is 404, an attacker can distinguish.

**5. Permission level.** \`assertAccess(UUID, Set<ProcessDataInclude>)\` is a single check — does it grant \`READ\`? If a future caller wants \`WRITE\` access (\`KILL\`, \`UPDATE_STATUS\`), do they get a separate method or a \`ResourceAccessLevel\` parameter? Worth adding the level argument now even if the current callers all default to \`READ\`, otherwise the third future caller will fork the helper.

Direction is right. Centralization is the win — the nits above are mostly about defending against the next refactor.

benbroadaway (approved):

PR #1320, opened 2026-05-06
webapp-plugin: fix caching policy

ibodrov:
- WebappFilter now serves webapp checksum resources with quoted ETag and Cache-Control: public, max-age=31536000, immutable.
  - If-None-Match now handles normal browser forms: lists, weak tags, and *, returning 304 on match.
  - HEAD is supported; non-GET/HEAD returns 405.


brig:
Looks like `/cfg.js` will now get `Cache-Control: public, max-age=2592000, **immutable**` from `WebappFilter`
Is that expected?

ibodrov:
> Looks like `/cfg.js` will now get `Cache-Control: public, max-age=2592000, **immutable**` from `WebappFilter` Is that expected?

No, that's unintentional. Thanks for catching it.

ibodrov:
Updated in 84ba5406e4f3eeaa2c37736aa2adddf55bd049cb. WebappFilter now only adds immutable Cache-Control for directly mapped checksum resources and preserves an existing Cache-Control header, so /cfg.js keeps the no-cache policy.

it/console/src/test/java/com/walmartlabs/concord/it/console/CachingHeadersIT.java
```diff
@@ -0,0 +1,170 @@
...
+
+        assertEquals(200, index.statusCode());
+        assertImmutableResource(index);
+        assertNotModified("/", header(index, "ETag"));
+
+        var resourcePath = findStaticResource(index.body());
+        var resource = get(resourcePath);
+
+        assertEquals(200, resource.statusCode());
+        assertImmutableResource(resource);
```
brig, line 71: Should `index.html` be `immutable`?

brig (approved):

PR #1334, opened 2026-05-25
concord-server: add wait condition for external process for suspend/resume

benbroadaway:
Adds an alternative, more reliable, option to standard suspend/resume API. 

Current issue: The standard suspend/resume pattern mostly works fine for one-to-one resumeEvent-to-externalWork relationships. Essentially, if someone uses a task that straightforwardly suspends and resumes from an external service callback, it works fine. But if they try to get too fancy with multiple resume events invoked at the same time (e.g. two threads calling the task which suspends, then there's a race condition if the two callbacks hit at the same time. This issue can be worked around by retrying the resume callback, but that adds noise and there are times when the resume-then-suspend-for-the-loser-to-retry situation takes longer than an external service will reasonably willing to retry. Further, there's no way to "timeout" any of the suspend/resume attempts--processes hit with the race condition can remain suspended indefinitely.

Solution: Add a new process wait condition which indicates the process is waiting for one or more specific external callbacks to be made. Very similar pattern overall, but with a little more control for the resuming side of things. 

There's some extra overhead to implementing this pattern, so it's advisable to use the "standard" suspend/resume pattern unless necessary.

The pattern:
1. Concord process tells <some other system> to do work and give's it a standard callback URL to call when it's done. The url contains a unique event id which for that specific remote work context (this is _not_ the resume event ID for the process)
    a. This can be repeated with another unique event id for each callback to be made.
2. Process creates an `EXTERNAL_EVENT` wait condition for each remote work event id, and a shared resume event id for the current thread of the process.
    a. This can be repeated across threads within a single process. Very importantly, each thread should use its own resume event ID AND one use one per thread/suspend call.
    b. Each wait condition can be customized with an `expiresAt` datetime value to ensure the process resumes in a timely manner even if the remote job is "lost"
3. Process suspends with the resume event ID(s).
4. Watchdog watches the wait conditions to check for expiration or 100% non-waiting for each wait condition with the same resume event ID for a given process ID.
5. External service(s) clear the wait conditions one at a time when they are done. They can optionally send variables which will eventually be added to the process state. Until then, they persist in the wait condition. (This is one reason a max-size is important).
6. When all of the wait conditions are satisfied for a given resume event ID, then the watchdog resumes the process execution with variables delivered under a name specified in the wait condition's `saveAs` field.
7. If multiple threads suspended for external event conditions, then the process may re-suspend to wait for the remaining thread's conditions to be met. If multiple thread's conditions were all met within the same watchdog run, then all threads will resume at the same time (no race condition).

Some things I'm not sure about yet:

* Naming is currently `WaitType.EXTERNAL_EVENT`. Makes sense, but that's the same name as an audit log event type. Too close to home?
* `expiresAt` default value. Currently set to 2 hours if not specified. Should default be configurable through server config or policy? Same for max value for the field (e.g. don't expire 3 allow expirations <too far into the future> (Currently set to 2 days).
* `process_wait_conditions` churn/bloat. This isn't a super-frequently called API, but just need to keep in mind what it does. Write (and rewrites for *N wait conditions) wait conditions, then re-write the conditions each time they're cleared by a callback which adds the resume vars. Probably wise to set a max-size for resume vars, and max number of waits allowed (the latter may be a little more complex than it initially seems).
* Regardless of size, this adds more quantity of wait conditions overall, and will extend the run time of the watchdog process for handling other wait conditions. Good news is that it doesn't add update queries when nothing changes to the waits between runs and doesn't require extra mutual-exclusivity queries like the `PROCESS_COMPLETION` waits. Mostly more listing calls added.
* cleaner way to indicate a wait condition expired when the process resumes?
* There may be some idempotency concerns. (network error-induced retry on create wait, but the first one actually succeeded, just didn't get the response delivered)


ibodrov (dismissed):
LGTM. Any plans to update `ProcessWaitList` to render `EXTERNAL_EVENT`s in the UI?

server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessExternalEventHandler.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public List<Result<ProcessExternalEventCondition>> processBatch(
+            List<WaitConditionItem<ProcessExternalEventCondition>> waits
+    ) {
+
+        // collect to grouping by process resume event. All waits for the same resume
+        // event must be completed before we can process them together and resume.
+        var byResumeEvent = waits.stream()
+                .collect(groupingBy(wait -> wait.waitCondition().resumeEvent()));
```
ibodrov, line 59: I think we need to group by `instanceId + resumeEvent` instead of just `resumeEvent`?

it/runtime-v2/src/test/java/com/walmartlabs/concord/it/runtime/v2/WaitConditionsIT.java
```diff
@@ -0,0 +1,153 @@
...
+    @BeforeAll
+    static void setup() {
+        processApi = new ProcessApi(concord.apiClient());
+    }
+
+    @Test
+    void testSingle() throws Exception {
+        var payload = new Payload()
+                .entryPoint("single")
+                .archive(resource("waitConditions"));
```
ibodrov, line 49: The PR is missing `waitConditions` folder

server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessExternalEventHandler.java
```diff
@@ -0,0 +1,146 @@
...
+            return Map.of();
+        }
+
+        var splits = saveAs.split("\\.");
+        if (splits.length > 10) { // that's suspiciously deep
+            logManager.warn(processKey, "External event condition 'saveAs' is too deep at {} levels. Resume variables will not be delivered.", splits.length);
+            return Map.of();
+        }
+
+        var variables = isExpired(condition) ? Map.of("isExpired", true) : condition.variables();
```
ibodrov: Do we need to check if waiting=false here? Might be unlucky race if someone calls clearExternalWaitCondition just before expiresAt and picked up by ProcessWaitWatchdog right after expiresAt.

Or maybe we do save variables unconditionally but also add `isExpired=true` to the map?

ibodrov (commented):
Some findings.

server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessExternalEventHandler.java
benbroadaway: Very good point. Makes me think just remove the injected `isExpired=true` from the resume variables.

If the race condition you described hits, then the result sent back to the process would be `valid=true` + `<valid callback-provided variables>`, so whatever handles the result will still have to add extra logic to suss out whether they got back what they expected when it appears it expired.

it/runtime-v2/src/test/java/com/walmartlabs/concord/it/runtime/v2/WaitConditionsIT.java
benbroadaway: Classic mistake ☹️

server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/WaitProcessExternalEventHandler.java
benbroadaway: There _shouldn't_ be more than one `instanceId` in the `process_wait_conditions` table, right? Each row contains all of the conditions for the given `instanceId`. Otherwise we'd have some wacky side effects with the existing wait processing. All wait conditions for a single process are handled [within the same batch item](https://github.com/walmartlabs/concord/blob/d53fcc07ac523fa6aade85b7436444f8897a3ea1/server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/ProcessWaitWatchdog.java#L115-L127). Which is why I probably need to size limit for the dynamic-length fields (e.g. `variables`, `externalEvent`, `reason`, `saveAs` to keep one process from blowing out memory if it has too many conditions or conditions that are too large (or both).

server/impl/src/main/java/com/walmartlabs/concord/server/process/waits/ProcessExternalEventCondition.java
```diff
@@ -0,0 +1,88 @@
...
+     * Indicates if the condition is still waiting for the event.
+     */
+    public abstract boolean waiting();
+
+    /**
+     * Optional payload of variables to set when the event is cleared.
+     */
+    @Nullable
+    @Value.Redacted
+    public abstract Map<String, Serializable> variables();
```
ibodrov, line 63: `@Value.Redacted` affects only `toString()` method.

server/impl/src/main/java/com/walmartlabs/concord/server/process/ProcessResource.java
```diff
@@ -942,14 +948,179 @@ public Response updateMetadata(@PathParam("id") UUID instanceId, Map<String, Obj
...
+        if (!updated.get()) {
+            return Response.status(Status.NOT_FOUND)
+                    .entity("No matching external event wait condition found or already cleared")
+                    .build();
         }
-        processWaitManager.addWait(processKey, condition);
+
+        // Persist the updated wait conditions
+        processWaitManager.tx(tx -> {
+            processWaitManager.setWait(tx, pk, waits, true, waitEntry.version());
```
ibodrov, line 1085: What should happen if setWait returns false here?

server/impl/src/main/java/com/walmartlabs/concord/server/process/ProcessResource.java
```diff
@@ -942,14 +948,179 @@ public Response updateMetadata(@PathParam("id") UUID instanceId, Map<String, Obj
...
+
+        if (condition.externalEvent() == null || condition.externalEvent().isBlank()) {
+            throw new ConcordApplicationException("External event name is required for external event wait condition", Status.BAD_REQUEST);
+        }
+
+        if (condition.externalEvent().length() > 128) {
+            throw new ConcordApplicationException("externalEvent value is too long", Status.BAD_REQUEST);
+        }
+
+        if (condition.resumeEvent().length() > 128) {
```
ibodrov, line 1019: Can `condition.resumeEvent()` be null here? 🤔

PR #1362, opened 2026-08-14
it: add agent-operator tests

ibodrov:


it/agent-operator/src/test/java/com/walmartlabs/concord/it/agentoperator/AgentOperatorIT.java
```diff
@@ -0,0 +1,381 @@
...
+
+        // pod traffic is masqueraded through the k3s node container, so the raw IP of the server
+        // container on the shared network is reachable from the pods (unlike Docker network aliases,
+        // which the pods can't resolve)
+        var serverIp = server.getContainerInfo().getNetworkSettings().getNetworks().values().iterator().next().getIpAddress();
+        assertNotNull(serverIp, "Can't determine the Concord Server container IP");
+        serverUrl = "http://" + serverIp + ":8001";
+        log.info("Concord Server URL for k3s pods: {}", serverUrl);
+
+        k3s = new K3sContainer(DockerImageName.parse(K3S_IMAGE))
```
benbroadaway:
```suggestion
        k3s = new K3sContainer(DockerImageName.parse(K3S_IMAGE).asCompatibleSubstituteFor("rancher/k3s"))
```

it/agent-operator/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
```
benbroadaway, line 65: Needs a `logback-test.xml` to suppress the debug output from `com.github.dockerjava.zerodep.shaded.org.apache.hc.client5` and other chatty libs.

benbroadaway (approved):

benbroadaway:
These tests were failing internally because the agent was hunting for `concord-targetplatform:2.44.1-SNAPSHOT` and couldn't connect to `repo.maven.apache.org`. I'm not really sure how the ITs passed in GH actions, but I fixed our internal situation by copying/caching `concord-targetplatform` in the image.

This isn't exactly the PR to do it, but we probably need to add an accommodation in the ITs (all around) to customize the `mvn.json` fed to the server/agent. Generally, the agent _shouldn't_ need to hit maven central, because it has already pulled the deps for the project defined by `concord-targetplatform`. However, looks like there's a few spots where a transitive dependency is...different...for some reason. e.g. after `concord-platform`, it was hunting for `org.apache.httpcomponents:httpclient:4.5.11`, even though we target [`httpclient:4.5.13`](https://github.com/walmartlabs/concord/blob/cae2452bb21b6ad652719fb32757a7180268411c/targetplatform/pom.xml#L66). Probably because we have a different target for [`httpmime:4.5.11`](https://github.com/walmartlabs/concord/blob/cae2452bb21b6ad652719fb32757a7180268411c/targetplatform/pom.xml#L68).

PR #1368, opened 2026-08-15
server: require 'agentWebsocket' permission for websockets

benbroadaway:
This will be a breaking change if an instance relies on the default (no user) token for agent websocket connections. You can disable the requirement with a server config. Not recommened though.

```
    websockets {
        requirePermission = false
    }
```

This makes the integration tests a little trickier to kick off. We may ought to rethink the default token (no user) approach. runtime-v1's ITs create a dedicated agent user and grant it permission, while the other ITs disable the check. There's a util method `ITUtils.createAgentUser()` to handle the creation, it just needs to be fed an api client at the right time. That can be reused for the other ITs, or it may be more appropriate for testcontainers-concord to handle that functionality.

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/agent/websocket/WebSocketCreator.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/agent/websocket/WebSocketCreator.java
@@ -26,7 +26,6 @@ import com.walmartlabs.concord.server.security.Permission;
 import com.walmartlabs.concord.server.security.UserSecurityContext;
 import com.walmartlabs.concord.server.security.apikey.ApiKeyDao;
 import com.walmartlabs.concord.server.security.apikey.ApiKeyEntry;
-import org.apache.shiro.authz.AuthorizationException;
 import org.eclipse.jetty.ee8.websocket.server.JettyServerUpgradeRequest;
 import org.eclipse.jetty.ee8.websocket.server.JettyServerUpgradeResponse;
 import org.eclipse.jetty.ee8.websocket.server.JettyWebSocketCreator;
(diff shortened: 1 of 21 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/agent/websocket/WebSocketCreator.java
```diff
@@ -72,12 +84,38 @@ public Object createWebSocket(JettyServerUpgradeRequest req, JettyServerUpgradeR
...
+            return null;
+        }
+
         String channelId = "ws-" + UUID.randomUUID();
         String agentId = req.getHeader(QueueClient.AGENT_ID);
         String userAgent = req.getHeader(QueueClient.AGENT_UA);
         return new WebSocketListener(channelManager, channelId, agentId, userAgent);
     }
 
+    private boolean hasPermission(UUID userId) {
```
ibodrov: userSecurityContext.runAs is a bit heavy to be wrapped into a simple-looking hasPermission, IMO. It rebinds the threadcontext, for example.

We should add something like

```java
// UserSecurityContext.java
public boolean isPermitted(UUID userId, Permission permission) {
    var user = userManager.get(userId).orElse(null);
    if (user == null) {
        return false;
    }

    var principal = new UserPrincipal(InternalRealm.REALM_NAME, user);
    var principals = new SimplePrincipalCollection(principal, InternalRealm.REALM_NAME);

    return securityManager.isPermitted(principals, permission.getKey());
}
```

benbroadaway: Agreed

eau-defemme (dismissed):

mtkunkel (approved):

PR #1374, opened 2026-08-21
concord-server: config option to enable template UI sources with scheme allow-list

benbroadaway:
Better control over template sources. Production instances should generally only allow templates from admin-configured aliases.

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/process/pipelines/processors/TemplateFilesProcessor.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/process/pipelines/processors/TemplateFilesProcessor.java
@@ -23,6 +23,7 @@ package com.walmartlabs.concord.server.process.pipelines.processors;
 import com.walmartlabs.concord.common.ZipService;
 import com.walmartlabs.concord.dependencymanager.DependencyManager;
 import com.walmartlabs.concord.sdk.Constants;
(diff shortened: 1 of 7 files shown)


eau-defemme (dismissed):

server/impl/src/main/java/com/walmartlabs/concord/server/process/pipelines/processors/TemplateFilesProcessor.java
```diff
@@ -92,6 +97,14 @@ private URI getUri(PartialProcessKey processKey, String template) throws URISynt
                 return getByAlias(processKey, template);
             }
 
+            if (!templatesConfiguration.isAllowUriSource()) {
+                throw new IllegalArgumentException("Templates must be configured through an alias.");
+            }
+
+            if (!templatesConfiguration.getAllowedUriScheme().contains(scheme)) {
```
brig: It looks like there's no way to specify "any scheme" in AllowedUriScheme, so the parameter is effectively required now. I don't think this will affect anyone outside Walmart, but maybe we should allow an empty list?

benbroadaway: I wouldn't recommend that for a prod instance, but fair enough to have the option if it's handy in test/isolation. Note: the scheme assertion is not applied to templates from an alias since an admin has to set that up in the first place.

brig (approved):

PR #1376, opened 2026-08-25
server: assert process access for decrypting project strings

eau-defemme:
otherwise the existence of a process allows you to decrypt whatever is going on in a project

### DIFF
--- a/server/impl/src/main/java/com/walmartlabs/concord/server/process/ProcessResource.java
+++ b/server/impl/src/main/java/com/walmartlabs/concord/server/process/ProcessResource.java
@@ -886,7 +886,7 @@ public class ProcessResource implements Resource {
                             @Parameter(schema = @Schema(type = "string", format = "byte"))
                             InputStream data) {
         ProcessEntry entry = assertProcess(PartialProcessKey.from(instanceId));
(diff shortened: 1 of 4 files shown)


server/impl/src/main/java/com/walmartlabs/concord/server/process/ProcessResource.java
```diff
@@ -886,6 +886,7 @@ public Response decrypt(@PathParam("id") UUID instanceId,
                             @Parameter(schema = @Schema(type = "string", format = "byte"))
                             InputStream data) {
         ProcessEntry entry = assertProcess(PartialProcessKey.from(instanceId));
+        assertProcessAccess(entry, ResourceAccessLevel.READER, "generic decryption");
```
benbroadaway: Let's give the description as something closer what we have in [other documentation](https://concord.walmartlabs.com/docs/plugins-v2/crypto.html#encrypting-and-decrypting-values). Like `project encrypted string`.

It's going to look a little funny in the resulting error message
> ```
>  The current user (the_user) doesn't have the necessary permissions to the
>  download project encrypted string
> ```

Actually it's not too bad, except i just noticed my own horrible ☹️ typo `the necessary permissions to the download` should be `the necessary permissions to download: `.

benbroadaway (approved):

PR #1377, opened 2026-09-03
server: assert access to org/project before listing triggers

eau-defemme:


server/impl/src/main/java/com/walmartlabs/concord/server/org/triggers/TriggerV2Resource.java
```diff
@@ -77,48 +79,49 @@ public List<TriggerEntry> list(@QueryParam("type") @ConcordKey String type,
...
-    }
-
-
-    private ProjectEntry assertProject(UUID projectId) {
-        if (projectId == null) {
-            throw new ValidationErrorsException("Invalid project ID or name");
+        // For org-wide queries (no specific project), filter to projects the caller can see.
+        // Admins and global readers/writers pass null to skip per-user filtering.
+        UUID currentUserId = null;
+        if (projectId == null && !(Roles.isAdmin() || Roles.isGlobalReader() || Roles.isGlobalWriter())) {
```
benbroadaway: `isGlobalWriter` should be omitted. There may be cases where some user is provisioned with write-only access (e.g. it just writes logs) but shouldn't be allowed to read globally. If it needs to read, then it should also be given the global reader permission appropriately.

benbroadaway (changes requested):

benbroadaway (approved):

